PR #5, opened 2018-06-21
Initial functions for resolve-tags plugin

priyawadhwa:
I decided to split up the plugin code into a few PRs so it wouldn't get too big, so this is two of the functions. The first recursively goes through the yaml and gets all tagged images, and the second resolves tags to digests. 

I also added a testutil package which we should be able to use for other unit tests.


cmd/kritis/kubectl/plugins/resolve/main.go
```diff
@@ -0,0 +1,102 @@
...
+		for _, v := range t {
+			images = append(images, recursiveGetTaggedImages(v)...)
+		}
+	}
+	return images
+}
+
+// resolveTagsToDigests resolves all images specified by tag to digest
+// It returns a map of the form [image:tag]:[image@sha256:digest]
+func resolveTagsToDigests(images []string) (map[string]string, error) {
```
dlorenc: in genereal we should try to keep main as small as possible. Could you put these in a different package? Maybe pkg/kritis/kubectl/plugins/resolve instead of cmd/

cmd/kritis/kubectl/plugins/resolve/main.go
```diff
@@ -0,0 +1,102 @@
...
+	//  2. Get tagged images using recursiveGetTaggedImages
+	// 	3. Resolve tagged images using resolveTagsToDigests
+	//	4. Write recursive function to replace images
+	//	5. Print to STDOUT
+	return nil
+}
+
+// recursiveGetTaggedImages recursively gets all images referenced by tags
+// instead of digests
+func recursiveGetTaggedImages(m interface{}) []string {
```
dlorenc: This should work, but it's worth considering doing this the same way kubectl does, which is essentially a more typed version of the same thing: https://github.com/kubernetes/kubernetes/blob/9eb0c3566808bb02de5ae6f43fa79da49718c864/pkg/kubectl/polymorphichelpers/updatepodspec.go#L33

priyawadhwa: So I don't think this can be further typed beyond `yaml.MapItem`, `yaml.MapSlice`, and `[]interface{}` (which is meant to represent `[]yaml.MapSlice`), unless I'm missing something?

cmd/kritis/kubectl/plugins/resolve/main.go
priyawadhwa: For sure, done!

PR #7, opened 2018-06-26
Add some Makefile rules to build things.

dlorenc:


Makefile
```diff
@@ -9,7 +9,37 @@
...
+
+%.exe: %
+	mv $< $@
+
+$(BUILD_DIR):
+	mkdir -p $(BUILD_DIR)
+
+ORG := github.com/grafeas
+PROJECT := kritis
+RESOLVER_PROJECT := resolver
```
priyawadhwa: In the plugin.yaml I named this binary [resolve-tags](https://github.com/grafeas/kritis/blob/master/cmd/kritis/kubectl/plugins/resolve/plugin.yaml#L3). I think resolver sounds good too, but they should probably match.

PR #22, opened 2018-06-27
Mock out image resolution.

dlorenc:
The tests started failing because these are real images and their digests can change :)


tejal29:
Thanks..  for mocking them out.

PR #23, opened 2018-06-27
Add flags to plugin and piece together functions

priyawadhwa:
I added the filename flag to the plugin.yaml and added cobra so that the binary can be used independently.


pkg/kritis/kubectl/plugins/resolve/resolve_test.go
```diff
@@ -223,6 +227,43 @@ func Test_recursiveReplaceImage(t *testing.T) {
...
+	testYaml := `apiVersion: v1
+kind: Pod
+metadata:
+  name: test
+spec:
+  containers:
+  - name: debian
+    image: %s`
+	initial := fmt.Sprintf(testYaml, "gcr.io/kaniko-project/executor:v0.1.0")
+	expected := fmt.Sprintf(testYaml, "gcr.io/kaniko-project/executor@sha256:501056bf52f3a96f151ccbeb028715330d5d5aa6647e7572ce6c6c55f91ab374")
```
dlorenc: do you want to mock this out and use a fixed value?

cmd/kritis/kubectl/plugins/resolve/cmd/root.go
```diff
@@ -0,0 +1,71 @@
...
+var (
+	files multiArg
+)
+
+func init() {
+	RootCmd.PersistentFlags().VarP(&files, "filename", "f", "Filename to resolve. Set it repeatedly for multiple filenames.")
+}
+
+var RootCmd = &cobra.Command{
+	Use: "resolve-tags",
```
aaron-prindle: currently running the binary with no -f flag succeeds but does nothing.  it might make sense to show help on run w/o -f flags or change the -f flags to be args

cmd/kritis/kubectl/plugins/resolve/cmd/root.go
```diff
@@ -0,0 +1,71 @@
...
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		return resolveFilepaths()
+	},
+	Run: func(cmd *cobra.Command, args []string) {
+		if err := resolve.Execute(files, os.Stdout); err != nil {
```
tejal29: instead of os.Stdout you can use, cmd.OutOrStdout().
This will helpful in testing the RootCmd.
You can add Tests for  end to end tests for RootCmd, 
```
	var OutputBuffer bytes.Buffer
	RootCmd.SetOutput(&OutputBuffer)
	RootCmd.SetArgs([]string{"--filename=<some temp file with temp data>"})
```

cmd/kritis/kubectl/plugins/resolve/cmd/root.go
```diff
@@ -0,0 +1,71 @@
...
+			fmt.Println(err)
```
tejal29: same here use RootCmd.Println which in this case will print to StdErr
https://github.com/spf13/cobra/blob/master/command.go#L1064
This will consistent with how cobra lib prints all errors.

cmd/kritis/kubectl/plugins/resolve/plugin.yaml
```diff
@@ -1,3 +1,8 @@
 name: "resolve-tags"
 shortDesc: "resolve images in kubernetes yamls from tags to digests"
-command: "./resolve-tags"
+command: "./resolve-tags $PWD -f=$KUBECTL_PLUGINS_LOCAL_FLAG_FILENAME"
```
tejal29: Sorry, i dont have much context, but from what i heard yesterday, you can't pass multiple args flags?
How will users use this plugin?
```
export KUBECTL_PLUGINS_LOCAL_FLAG_FILENAME=myfile.yaml
kubectl resolve-tags
```
or
```
kubectl resolve-tags -f myfile.yaml
```

pkg/kritis/kubectl/plugins/resolve/resolve.go
```diff
@@ -126,3 +144,9 @@ func recursiveReplaceImage(i interface{}, replacements map[string]string) interf
 	}
 	return nil
 }
+
+// prints the final replaced kubernetes manifest to STDOUT
+func print(mfst []byte, writer io.Writer) {
```
tejal29: Change this description. Print to given writer.

tejal29: fmt.Fprintln to write to same writer.?

cmd/kritis/kubectl/plugins/resolve/cmd/root.go
```diff
@@ -0,0 +1,71 @@
...
+			os.Exit(1)
+		}
+	},
+}
+
+func resolveFilepaths() error {
```
tejal29: Please add tests for this function
You can also test for RootCmd. Though there isnt a lot of logic here to test, it will be good smoke test. 
You also add a test to set multiple args and make sure they get parsed correctly.
```
	var OutputBuffer bytes.Buffer
	RootCmd.SetOutput(&OutputBuffer)
	RootCmd.SetArgs([]string{"--filename=<some temp file with temp data>"})
        err := RootCmd.Execute() //This parse the args, run prepersisted run etc.
       //Check error
      // Check output buffer.
```

cmd/kritis/kubectl/plugins/resolve/plugin.yaml
tejal29: Can we also add a short README with usage in plugins/resolve/README ?

pkg/kritis/kubectl/plugins/resolve/resolve_test.go
priyawadhwa: I sort of treated this like an integration test and chose an image that we probably won't change just to make sure the entire flow worked. Would it make more sense to do that somewhere else?

tejal29: Yeah we dont want to rely on already existing image also used for other purpose.
If you want to add an integration test, we can create a new image specific for this purpose and push it to a project dedicated for integration testing.
I am going to create "kritis-int-test" project and set it up for the team.
You can push a new image to this project. 
Does that make sense?

priyawadhwa: Yup sounds good!

priyawadhwa:
Yah, I interpreted KUBECTL_PLUGINS_LOCAL_FLAG_FILENAME being set as an indication that we were running via plugin. I'll add a quick note in the description so it's clear.

Issue #10, opened 2018-06-26
Add Background Go job for watching pods.


tejal29:
This is background go process which will use its own clock to sleep for an hour and check pods after every hour.

tejal29:
Skeleton Added #34 
Image Policy Validation Logic Added in #36

tejal29:
@priyawadhwa  this should be addressed in #53 ?

priyawadhwa:
Yah, all of the validation checks should be done.  The only thing left to do is add checks for AttestationAuthorities once the lib is merged.

PR #34, opened 2018-07-02
Setup the skeleton of the cron job.

dlorenc:
Signed-off-by: dlorenc <dlorenc@google.com>

### DIFF
--- a/pkg/kritis/cron/cron_test.go
+++ b/pkg/kritis/cron/cron_test.go
@@ -1,0 +1,52 @@
+/*
+Copyright 2018 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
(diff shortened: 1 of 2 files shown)


dlorenc:
The Start() function here can be called in a goroutine to start the cron server inside the main of the overall server.

dlorenc:
Ref #10

pkg/kritis/cron/cron_test.go
```diff
@@ -0,0 +1,53 @@
...
+		fmt.Println("HERE")
+		checked = true
+	}
+	defer func() {
+		podChecker = originalChecker
+	}()
+
+	// Set a deadline of longer than the check interval
+	ctx := context.Background()
+	c, _ := context.WithDeadline(ctx, time.Now().Add(10*checkInterval))
```
tejal29: check error.

pkg/kritis/cron/cron_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	// Speed up the checks for testing.
+	checkInterval = 2 * time.Millisecond
+
+	checked := false
+
+	// Mock the check function and reset after the test.
+	originalChecker := podChecker
+	podChecker = func() {
+		fmt.Println("HERE")
```
tejal29: Can remove this.

pkg/kritis/cron/cron_test.go
dlorenc: It wasn't actually an error, but a manual cancel func. I named it and called it in a defer to be more clear.

pkg/kritis/cron/cron_test.go
dlorenc: Done.

PR #36, opened 2018-07-03
Add pod checking to the cron job.

dlorenc:


pkg/kritis/cron/cron.go
```diff
@@ -19,32 +19,96 @@ package cron
...
+	"k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/sirupsen/logrus"
 )
 
 // For testing
 var (
-	podChecker    = CheckPods
 	checkInterval = 1 * time.Minute
```
priyawadhwa: I think this should be changed to every hour

pkg/kritis/cron/cron.go
```diff
@@ -19,32 +19,96 @@ package cron
...
+	ViolationStrategy violation.Strategy
+}
+
+var (
+	defaultViolationStrategy = &violation.LoggingStrategy{}
 )
 
+func NewCronConfig(cs *kubernetes.Clientset, ca containeranalysis.ContainerAnalysis) *Config {
+	pl := func(ns string) (*v1.PodList, error) {
+		return cs.CoreV1().Pods(ns).List(metav1.ListOptions{})
```
priyawadhwa: We have this [function](https://github.com/grafeas/kritis/blob/master/pkg/kritis/pods/pods.go#L36) `Pods(namespace string) ([]corev1.Pod, error)` in the pods package, which we could use here?

pkg/kritis/cron/cron.go
```diff
@@ -19,32 +19,96 @@ package cron
...
-func Start(ctx context.Context) {
+func Start(ctx context.Context, cfg Config) {
 	c := time.NewTicker(checkInterval)
 	done := ctx.Done()
 
 	for {
 		select {
 		case <-c.C:
-			podChecker()
+			if err := podChecker(cfg, nil); err != nil {
```
priyawadhwa: We can get the ImageSecurityPolicies to pass in to `podChecker` from [here](https://github.com/grafeas/kritis/blob/master/pkg/kritis/crd/securitypolicy/securitypolicy.go#L31)

pkg/kritis/cron/cron.go
```diff
@@ -19,32 +19,96 @@ package cron
...
-	// Fetch ISP crds
-	// Validate against all running pods
+func CheckPods(cfg Config, isps []v1beta1.ImageSecurityPolicy) error {
+	for _, isp := range isps {
+		pods, err := cfg.PodLister(isp.Namespace)
+		if err != nil {
+			return err
+		}
+		for _, p := range pods.Items {
+			for _, c := range p.Spec.Containers {
```
priyawadhwa: We have [Images](https://github.com/grafeas/kritis/blob/master/pkg/kritis/pods/pods.go#L46) to get all images in a pod (it also looks at InitContainers)

pkg/kritis/cron/cron_test.go
```diff
@@ -44,9 +54,143 @@ func TestStartCancels(t *testing.T) {
...
 		t.Fatalf("check function not called")
 	}
 }
+
+type imageViolations struct {
+	imageMap map[string]bool
+}
+
+func (iv *imageViolations) violationChecker(image string, isp v1beta1.ImageSecurityPolicy) ([]metadata.Vulnerability, error) {
+	if iv == nil {
```
priyawadhwa: Is this check necessary?

pkg/kritis/cron/cron.go
```diff
@@ -19,32 +19,100 @@ package cron
...
 var (
+	checkInterval = 1 * time.Hour
 	podChecker    = CheckPods
-	checkInterval = 1 * time.Minute
 )
 
+// For testing.
+type podLister func(string) ([]corev1.Pod, error)
+type violationChecker func(string, v1beta1.ImageSecurityPolicy) ([]metadata.Vulnerability, error)
+type violationHandler func(string, *corev1.Pod, []metadata.Vulnerability) error
```
priyawadhwa: I don't think violationHandler is being used anywhere?

Issue #2, opened 2018-06-21
kubectl plugin to resolve tags to digests


priyawadhwa:

dlorenc:
Some more details/ideas around the surface here:

The syntax for files should roughly match that of kubectl:

```shell
  -f, --filename=[]: Filename, directory, or URL to files that contains the configuration to apply
```

When run with no arguments, defaults to stdin.

We should have another mode that uses the `KUBECTL_PLUGINS_CALLER` env var to actually then invoke kubectl after. Something like:

```shell
kubectl plugin resolve-tags apply -f foo.yaml bar.yaml
```

Would do the resolution then immediately call apply, removing the need for people to do:

```shell
kubectl apply $(kubectl plugin resolve-tags -f foo.yaml bar.yaml)
```

tejal29:
is this done?

dlorenc:
I think it still needs some work to be finished up.

PR #39, opened 2018-07-06
Add --apply flag to run kubectl apply

priyawadhwa:
I changed `Execute()` in the `resolve` package to return a map of [file name]:[new contents]. If the --apply flag is set, the plugin will save the new contents to a commands stdin and then execute the `kubectl apply -f` command.

Should finish #2

### DIFF
--- a/cmd/kritis/kubectl/plugins/resolve/README.md
+++ b/cmd/kritis/kubectl/plugins/resolve/README.md
@@ -19,6 +19,11 @@ You can then run the plugin:
 kubectl plugin resolve-tags -f <path to file>
 ```
 
+To apply subsitutions using `kubectl apply -f`, you can run:
+```
+kubectl plugin resolve-tags -f <path to file> -a true
+```
+
 ## resolve-tags binary
 
 resolve-tags can also be run as a binary. 
@@ -35,4 +40,8 @@ You can then run the binary:
 
 ```
 ./out/resolve-tags -f <path to file> -f <path to another file>
+```
+To apply subsitutions using `kubectl apply -f`, you can run:
+```
+kubectl plugin resolve-tags -f <path to file> -a
 ```
--- a/cmd/kritis/kubectl/plugins/resolve/cmd/root.go
+++ b/cmd/kritis/kubectl/plugins/resolve/cmd/root.go
@@ -20,21 +20,28 @@ import (
 	"fmt"
 	"github.com/grafeas/kritis/pkg/kritis/kubectl/plugins/resolve"
 	"github.com/spf13/cobra"
+	"io"
 	"os"
+	"os/exec"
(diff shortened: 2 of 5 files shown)


cmd/kritis/kubectl/plugins/resolve/README.md
```diff
@@ -19,6 +19,11 @@ You can then run the plugin:
 kubectl plugin resolve-tags -f <path to file>
 ```
 
+To apply subsitutions using `kubectl apply -f`, you can run:
+```
+kubectl plugin resolve-tags -f <path to file> -a true
```
tejal29: Add an explanation for what  -a stands  before this line or use full name?
As soon as i read it, i thought it stood for --append.

priyawadhwa: For sure, done!

cmd/kritis/kubectl/plugins/resolve/cmd/root.go
```diff
@@ -75,3 +92,38 @@ func resolveFilepaths() error {
...
+func print(substitutes map[string]string, writer io.Writer) {
+	for file, contents := range substitutes {
+		fmt.Fprintln(writer, fmt.Sprintf("---%s---", file))
+		fmt.Fprintf(writer, contents)
+		fmt.Fprintln(writer)
+	}
+}
+
+func applyChanges(substitutes map[string]string, writer io.Writer) error {
+	kubectl := os.Getenv("KUBECTL_PLUGINS_CALLER")
```
tejal29: Is there a reason we need to do this? Can you add some comments here?

cmd/kritis/kubectl/plugins/resolve/cmd/root.go
```diff
@@ -46,16 +53,26 @@ var RootCmd = &cobra.Command{
...
 		   it will override any files passed in.`,
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		resolveApply()
 		return resolveFilepaths()
 	},
 	Run: func(cmd *cobra.Command, args []string) {
-		if err := resolve.Execute(files, cmd.OutOrStdout()); err != nil {
+		substitutes, err := resolve.Execute(files)
+		if err != nil {
+			cmd.Println(err)
```
tejal29: Refactor into ExitIfErr so you dont have to repeat the lines.

cmd/kritis/kubectl/plugins/resolve/cmd/root.go
priyawadhwa: This gives us the full path to the kubectl binary, which I figured would be better to use if we could get it. Otherwise, we assume it's on the PATH

PR #41, opened 2018-07-06
Return interface as default when replacing tags and unit test

### LINKED ISSUE #40: resolve-tags plugin doesn't correctly handle arrays
Found this bug in the resolve-tags plugin trying to resolve the generic kaniko template
Given this file:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: kaniko
spec:
  containers:
  - name: kaniko
    image: gcr.io/kaniko-project/executor:latest
    args: ["--dockerfile=<path to Dockerfile>",
            "--bucket=<GCS bucket>",
            "--destination=<gcr.io/$PROJECT/$IMAGE:$TAG>"]
    volumeMounts:
      - name: kaniko-secret
        mountPath: /secret
    env:
      - name: GOOGLE_APPLICATION_CREDENTIALS
        value: /secret/kaniko-secret.json
  restartPolicy: Never
  volumes:
    - name: kaniko-secret
      secret:
        secretName: kaniko-secret
```

The plugin printed:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: kaniko
spec:
  containers:
  - name: kaniko
    image: gcr.io/kaniko-project/executor@sha256:501056bf52f3a96f151ccbeb028715330d5d5aa6647e7572ce6c6c55f91ab374
    args:
    - null
    - null
    - null
    volumeMounts:
    - name: kaniko-secret
      mountPath: /secret
    env:
    - name: GOOGLE_APPLICATION_CREDENTIALS
      value: /secret/kaniko-secret.json
  restartPolicy: Never
  volumes:
  - name: kaniko-secret
    secret:
      secretName: kaniko-secret
```

priyawadhwa:
Fixes #40 

Arrays of strings, like in the kaniko example, were being typed as an interface{}, so the default nil was being returned. Instead, I changed the default to return the interface (which should always be a string)

### DIFF
--- a/pkg/kritis/kubectl/plugins/resolve/resolve_test.go
+++ b/pkg/kritis/kubectl/plugins/resolve/resolve_test.go
@@ -31,6 +31,9 @@ spec:
   containers:
   - name: docker
     image: golang:1.10
+    args: ["--arg1=<first>",
+	       "--arg2=<second>",
+	       "--arg3=<third>"]
 `
 
 var testYaml2 = `apiVersion: v1
--- a/pkg/kritis/kubectl/plugins/resolve/resolve.go
+++ b/pkg/kritis/kubectl/plugins/resolve/resolve.go
@@ -141,8 +141,9 @@ func recursiveReplaceImage(i interface{}, replacements map[string]string) interf
 			t[index] = replacedMapSlice
 		}
 		return t
+	default:
+		return t
 	}
-	return nil
 }
 
 // prints the final replaced kubernetes manifest to given writer


pkg/kritis/kubectl/plugins/resolve/resolve_test.go
```diff
@@ -31,6 +31,9 @@ spec:
   containers:
   - name: docker
     image: golang:1.10
+    args: ["--arg1=<first>",
+	       "--arg2=<second>",
```
tejal29: Fix format.

PR #45, opened 2018-07-10
Added validation check to admission webhook and unit tests

### LINKED ISSUE #8: Add Mutating Admission Validator Webhook.
This Admission Webhook will
1. watch for Image policy CRD.
2. use drydock api to fetch vuln occurrences (depends on #3 )
3. Implements the flow in DD shared internally

priyawadhwa:
Currently the admission webhook accepts everything, so I added a check for the "breakglass" annotation and validation against ImageSecurityPolicies.

Partially fixes #8, although we still need to add all the AttestationAuthority stuff

### DIFF
--- a/pkg/kritis/admission/admission.go
+++ b/pkg/kritis/admission/admission.go
@@ -18,45 +18,147 @@ package admission
 
 import (
 	"encoding/json"
+	"fmt"
+	"github.com/grafeas/kritis/pkg/kritis/admission/constants"
+	kritisv1beta1 "github.com/grafeas/kritis/pkg/kritis/apis/kritis/v1beta1"
+	"github.com/grafeas/kritis/pkg/kritis/crd/securitypolicy"
+	"github.com/grafeas/kritis/pkg/kritis/kubectl/plugins/resolve"
+	"github.com/grafeas/kritis/pkg/kritis/metadata"
+	"github.com/grafeas/kritis/pkg/kritis/metadata/containeranalysis"
+	"github.com/grafeas/kritis/pkg/kritis/pods"
+	"io/ioutil"
+	"k8s.io/api/admission/v1beta1"
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"log"
 	"net/http"
-
-	"github.com/grafeas/kritis/pkg/kritis/admission/constants"
-	"k8s.io/api/admission/v1beta1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// This is very basic admission controller which allows all pods.
+type config struct {
+	retrievePod           func(r *http.Request) (*v1.Pod, error)
+	imagesecuritypolicies func(namespace string) ([]kritisv1beta1.ImageSecurityPolicy, error)
+	validate              func(isp kritisv1beta1.ImageSecurityPolicy, project, image string, client metadata.MetadataFetcher) ([]metadata.Vulnerability, error)
+}
+
+var (
+	// For testing
+	admissionConfig = config{
+		retrievePod:           pod,
+		imagesecuritypolicies: securitypolicy.ImageSecurityPolicies,
+		validate:              securitypolicy.ValidateImageSecurityPolicy,
+	}
+)
+
+// This admission controller looks for the breakglass annotation
+// If one is not found, it validates against image security policies
+// TODO: Check for attestations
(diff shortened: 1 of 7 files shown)


pkg/kritis/admission/admission.go
```diff
@@ -18,45 +18,149 @@ package admission
...
 	"k8s.io/api/admission/v1beta1"
+	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"log"
+	"net/http"
 )
 
-// This is very basic admission controller which allows all pods.
+type config struct {
+	retrievePod           func(r *http.Request) (*v1.Pod, error)
```
dlorenc: These would be easier to mock if you take the pod name instead of the http request

pkg/kritis/admission/admission.go
```diff
@@ -18,45 +18,149 @@ package admission
...
+	imagesecuritypolicies func(namespace string) ([]kritisv1beta1.ImageSecurityPolicy, error)
+	validate              func(isp kritisv1beta1.ImageSecurityPolicy, project, image string, client metadata.MetadataFetcher) ([]metadata.Vulnerability, error)
+}
+
+var (
+	// For testing
+	admissionConfig = config{
+		retrievePod:           pod,
+		metadata:              metadataClient,
+		imagesecuritypolicies: securitypolicy.ImageSecurityPolicies,
```
dlorenc: could you name these as things like "fetchImageSecurityPolicies", as verbs?

pkg/kritis/admission/admission.go
```diff
@@ -18,45 +18,149 @@ package admission
...
+	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"log"
+	"net/http"
 )
 
-// This is very basic admission controller which allows all pods.
+type config struct {
+	retrievePod           func(r *http.Request) (*v1.Pod, error)
+	metadata              func() (metadata.MetadataFetcher, error)
```
dlorenc: can we use the interface here directly, instead of mocking out the function?

pkg/kritis/admission/admission.go
```diff
@@ -18,45 +18,149 @@ package admission
...
+	for _, image := range images {
+		if !resolve.FullyQualifiedImage(image) {
+			returnStatus(constants.FailureStatus, fmt.Sprintf("%s is not a fully qualified image", image), w)
+			return
+		}
+	}
+	// Third, validate images in the pod against ImageSecurityPolicies in the same namespace
+	isps, err := admissionConfig.imagesecuritypolicies(pod.Namespace)
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
```
tejal29: Can we  log the error here?
You can simply do log.Printf()

pkg/kritis/admission/admission.go
```diff
@@ -18,45 +18,149 @@ package admission
...
+		return
+	}
+	// get the client we will get vulnz from
+	metadataClient, err := admissionConfig.metadata()
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
```
tejal29: Log error.

pkg/kritis/admission/admission.go
```diff
@@ -18,45 +18,149 @@ package admission
...
+				return
+			}
+			if len(violations) != 0 {
+				//  TODO: Check AttestationAuthorities to see if the image is verified
+				returnStatus(constants.FailureStatus, fmt.Sprintf("found violations in %s", image), w)
+				return
+			}
+		}
+	}
+	// At this point, we can return a success status
```
tejal29: Add a TODO to add attestations

pkg/kritis/admission/admission.go
priyawadhwa: yah that's what I tried initially, but I think when fetching the metadata client it looks for some google auth so it doesn't work with travis

pkg/kritis/admission/admission.go
priyawadhwa: sorry, I'm not sure how that would change mocking out the function? since I wouldn't be using the http request or the pod name, and just returning a pod.

pkg/kritis/admission/admission.go
priyawadhwa: done

aaron-prindle:
it might be useful to add some examples .yaml files we can point to in the README.md

priyawadhwa:
for sure, added in #51 !

PR #51, opened 2018-07-12
Sample image security policy yamls

priyawadhwa:


artifacts/examples/image-security-policy-crd.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
tejal29: shall we move the definition one level up of example?

priyawadhwa: done

PR #52, opened 2018-07-12
Add Attestation lib to create a pgp signature and Verify it.

### LINKED ISSUE #11: Write a lib to create Attestations for an image.
- Write a lib to create Attestation Notes from Attestion Authority CRD.
-  The GPG_SIGNATURE will be computed over the Red Hat’s Atomic Host attestation signature format. We will be using only the required fields as mentioned in BinAuth Signature Format
```
{
    "critical": {
        "identity": {
            "docker-reference": "gcr.io/tejaldesai-personal/image-whitelist-server"
        },
        "image": {
            "docker-manifest-digest": "2a9906e7e50e7eca27c6ac6d1d32ae16ea4da813c593d21fc20d1cc534182f4d"
        },
        "type": "atomic container signature"
    },

```

- Create an AttestationOccurence  using the lib (#3)

- Write a function to verify Attestations.
Rules:
  - In order to admit an image which has attestations but fails image policy, we need to make sure all attestation are valid.
  - In order to verify a signature, we need to follow the following steps.
  - Fetch the AttestationAuthority CRD from AttestationAuthority AttestationOccurence.NoteName
  - Make sure the AttestationOccurence.attestation.pgpSignedAttestation.pgpKeyId is same as defined in the AttestationAuthority.spec.privateKeySecretName. If they are not same, the user changed the key for this AttestationAuthority. Skip verifying this Occurrence. Look for other attestations with same key.
  - Decrypt the pgpSignature using the Kubernetes Secret and Verify we get the same Container name 
 and Container digest from the atomic host attestation signature format.

tejal29:
Fixes #11 

This library, add logics to attest and verify a message using pgp keys.

TODO: Add more testing!


pkg/kritis/attestation/attestation.go
```diff
@@ -0,0 +1,206 @@
...
+		return err
+	}
+
+	sig, ok := pkt.(*packet.Signature)
+	if !ok {
+		return fmt.Errorf("Not a valid signature")
+	}
+
+	hash := sig.Hash.New()
+	io.Copy(hash, bytes.NewReader(b.Bytes))
```
dlorenc: nit, you may need to check this error.

pkg/kritis/attestation/attestation.go
```diff
@@ -0,0 +1,206 @@
...
+
+	err = key.VerifySignature(hash, sig)
```
dlorenc: nit: just return key.VerifySignature

pkg/kritis/attestation/attestation.go
```diff
@@ -0,0 +1,206 @@
...
+			Hash:         pgpConfig.Hash(),
+			IsPrimaryId:  &isPrimaryId,
+			FlagsValid:   true,
+			FlagSign:     true,
+			FlagCertify:  true,
+			IssuerKeyId:  &e.PrimaryKey.KeyId,
+		},
+	}
+
+	keyLifetimeSecs := uint32(86400 * 365)
```
dlorenc: should we make these constants?

pkg/kritis/attestation/attestation.go
tejal29: done

pkg/kritis/attestation/attestation.go
tejal29: done

tejal29:
DO NOT Review Yet!!!

tejal29:
@wilfried Please review!

pkg/kritis/attestation/attestation.go
```diff
@@ -0,0 +1,164 @@
...
+		return fmt.Errorf("Signature could not be verified. Got: %q, Want: %q", plaintext, message)
+	}
+	return nil
+}
+
+// AttestMessage attests the message using the given public and private key.
+// pubKeyEnc: Base64 Encoded Public Key
+// privKeyEnc: Base64 Decoded Private Key
+// message: Message to attest
+func AttestMessage(pubKeyEnc string, privKeyEnc string, message string) (string, error) {
```
wteiken: Nit/Optional: I would probably name this CreateImageAttestation to be aligned with VerifyImageAttestation.  Or name the other one also "VerifyMessageAttestation" as there really is no image specific code in here.

Could you point me to the code where this is called?

tejal29: it should be called here https://github.com/grafeas/kritis/blob/master/pkg/kritis/admission/admission.go#L108 where the todo is mentioned. 
We will first build a message of type  Red Hat’s Atomic Host attestation signature

tejal29: Added more details TODO here https://github.com/grafeas/kritis/pull/52/commits/b8cdbe9f336bf43c83ed2b3c833928b0895b7598

PR #53, opened 2018-07-12
Add AnnotationStrategy to annotate pods and refactor security policy

### LINKED ISSUE #47: Add ViolationStrategy to add labels and annotations.
Right now, we only log violated pods.
https://github.com/dlorenc/kritis/blob/5750fc4d74d1c6dd51d15f57002ad9a9b1bc26b0/pkg/kritis/violation/strategy.go#L31

- [ ] Add a Strategy to add labels and annotations
- [ ] Switch Default Logging Strategy to the newly added one.

priyawadhwa:
I refactored security policy to return a SecurityPolicyViolation instead of a list of the vulnz in violation. The SecurityPolicyViolation contains a vulnerability, along with why it failed the check (`Violation`) and a more descriptive `Reason` why it failed. I check the Violations when formatting the labels to add, and include all the Reasons in the annotations.

Fixes #47


pkg/kritis/crd/securitypolicy/securitypolicy.go
```diff
@@ -48,17 +49,24 @@ func ImageSecurityPolicies() ([]v1beta1.ImageSecurityPolicy, error) {
 
 // ValidateImageSecurityPolicy checks if an image satisfies ISP requirements
 // It returns a list of vulnerabilites that don't pass
-func ValidateImageSecurityPolicy(isp v1beta1.ImageSecurityPolicy, project, image string, client metadata.MetadataFetcher) ([]metadata.Vulnerability, error) {
+func ValidateImageSecurityPolicy(isp v1beta1.ImageSecurityPolicy, project, image string, client metadata.MetadataFetcher) ([]SecurityPolicyViolation, error) {
```
dlorenc: hmm, as a further refactor would it make sense to pass in a name.Digest https://github.com/google/go-containerregistry/blob/master/pkg/name/digest.go#L30 instead of a string? That way we don't have to keep checking that things are fully resolved everywhere.

pkg/kritis/crd/securitypolicy/violation.go
```diff
@@ -0,0 +1,62 @@
...
+package securitypolicy
+
+import (
+	"fmt"
+	"github.com/grafeas/kritis/pkg/kritis/apis/kritis/v1beta1"
+	"github.com/grafeas/kritis/pkg/kritis/constants"
+	"github.com/grafeas/kritis/pkg/kritis/metadata"
+)
+
+type violation int
```
dlorenc: Hmm, maybe this would be a bit nicer as `type violation string`, then your functions below can cast the strings into the `type violation`, that way we'd get a bit more type safety.

pkg/kritis/crd/securitypolicy/violation.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// UnqualifiedImageViolationReason returns a detailed reason if the image is unqualified
+func UnqualifiedImageViolationReason(image string) string {
+	return fmt.Sprintf("%s is not a fully qualified image", image)
+}
+
+// FixesAvailableViolationReason returns a detailed reason if a CVE doesn't have a fix available
+func FixesNotAvailableViolationReason(image string, vulnz metadata.Vulnerability) string {
+	return fmt.Sprintf("found CVE %s in %s which doesn't have fixes available", vulnz.CVE, image)
```
tejal29: You need to reverse of this. I was also confused first. but the way this should be implemented is:

For Policy Spec, with OnlyFixesNotAvailable as false (default true) we allow images with vulnerabilities which have fixes out. If the fix is out, we want to encourage users to rebuild image and deploy with that image. Hence the default is set to true.
So the reason should be `found CVE %s in %s which has fixes available`
```
kind: ImageSecurityPolicy
metadata:
  name: my-isp
  namespace: default
spec:
  imageWhitelist:
  - gcr.io/my/image
  packageVulnerabilityRequirements:
    maximumSeverity: HIGH
    onlyFixesNotAvailable: false
 
```

pkg/kritis/violation/strategy.go
```diff
@@ -16,29 +16,65 @@ limitations under the License.
...
 }
 
+// AnnotationStrategy adds "InvalidImageSecPolicy" as a label and an annotation, with detailed info in the annotations
+type AnnotationStrategy struct {
+}
+
+func (a *AnnotationStrategy) HandleViolation(image string, pod *v1.Pod, violations []securitypolicy.SecurityPolicyViolation) error {
+	// First, remove "InvalidImageSecPolicy" label/annotation in case it doesn't apply anymore
+	deletionKeys := []string{constants.InvalidImageSecPolicyLabel}
+	if err := pods.DeleteLabelsAndAnnotations(*pod, deletionKeys, deletionKeys); err != nil {
```
tejal29: does it return error when label and annotations does not exist?

tejal29: and you are trying to delete the annotations and label

pkg/kritis/crd/securitypolicy/securitypolicy.go
priyawadhwa: So right now we do the check in two places:

1. For the cron job, I like keeping the check here in ValidateImageSecurityPolicy because we also get information about CVE violations, and all of that gets added as an annotation to the pod. If we did the check before calling this, we wouldn't get that.

2. In the webhook we do it before we even check against security policies because it results in an automatic denial

pkg/kritis/crd/securitypolicy/violation.go
priyawadhwa: Thanks for explaining this! :)

pkg/kritis/violation/strategy.go
priyawadhwa: Yah, I delete any previous annotations/labels in case they no longer apply, and then add the new ones. If the annotation or label doesn't exist, it won't return an error.

pkg/kritis/crd/securitypolicy/violation.go
```diff
@@ -0,0 +1,62 @@
...
+package securitypolicy
+
+import (
+	"fmt"
+	"github.com/grafeas/kritis/pkg/kritis/apis/kritis/v1beta1"
+	"github.com/grafeas/kritis/pkg/kritis/constants"
+	"github.com/grafeas/kritis/pkg/kritis/metadata"
+)
+
+type violation string
```
dlorenc: Ah, here I actually meant to make the UnqualifiedImageViolationReason function return a "violation" string, instead of using this type as the enum.

pkg/kritis/violation/strategy.go
tejal29: thanks.

pkg/kritis/crd/securitypolicy/violation.go
priyawadhwa: Oh, thanks for clarifying. I'll return the "violation" string and switch the enum back to ints, is that correct?

dlorenc: right, thanks! One last nit: since these are Public methods, you should make the Violation type Public as well.

PR #61, opened 2018-07-13
make caBundle configurable via helm --set

aaron-prindle:


aaron-prindle:
with this users can run:
helm install ./kritis-charts --set caBundle=$(kubectl get csr tls-webhook-secret-cert -o jsonpath='{.status.certificate}' )

kritis-charts/templates/kritis-server-webhook.yaml
```diff
@@ -16,7 +16,6 @@ webhooks:
           - pods
     failurePolicy: Fail
     clientConfig:
-      caBundle: # TODO: Add script to inject caBundle here
       service:
```
priyawadhwa: I think we need to add `caBundle: {{ .Values.caBundle}}` here?

aaron-prindle: yea sorry, missed this switching things over. added now

PR #64, opened 2018-07-16
namespace labels/annotations with kritis.grafeas.io

### LINKED ISSUE #63: namespace all the labels/annotations
We should namespace all the labels and annotations with something like:

kritis.grafeas.io/<key>

priyawadhwa:
Should fix #63

### DIFF
--- a/pkg/kritis/admission/admission.go
+++ b/pkg/kritis/admission/admission.go
@@ -20,8 +20,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"github.com/grafeas/kritis/pkg/kritis/admission/constants"
-	"github.com/grafeas/kritis/pkg/kritis/apis/kritis"
 	kritisv1beta1 "github.com/grafeas/kritis/pkg/kritis/apis/kritis/v1beta1"
+	kritisconstants "github.com/grafeas/kritis/pkg/kritis/constants"
 	"github.com/grafeas/kritis/pkg/kritis/crd/securitypolicy"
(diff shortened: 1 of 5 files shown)


pkg/kritis/violation/strategy.go
```diff
@@ -63,8 +64,8 @@ func (a *AnnotationStrategy) HandleViolation(image string, pod *v1.Pod, violatio
 		}
 		annotationValue += string(v.Reason) + "\n"
 	}
-	labels := map[string]string{constants.InvalidImageSecPolicyLabel: labelValue}
-	annotations := map[string]string{constants.InvalidImageSecPolicyLabel: annotationValue}
+	labels := map[string]string{kritis.GroupName: labelValue}
```
tejal29: The label name should be "{grpname}.{LabelName}".format(kritis.GroupName, constatants.InvalidImageSecPolicy
Label)

priyawadhwa: done

pkg/kritis/admission/admission.go
```diff
@@ -132,7 +132,8 @@ func checkBreakglass(pod *v1.Pod) bool {
 	if annotations == nil {
 		return false
 	}
-	_, ok := annotations[constants.BREAKGLASS]
+	breakglass := fmt.Sprintf("%s/%s", kritis.GroupName, constants.BREAKGLASS)
```
dlorenc: Might be nice to make all of these labels be constants themselves, declared somewhere globally.

priyawadhwa: done

PR #70, opened 2018-07-16
Add global image whitelist

### LINKED ISSUE #48: Add a Global Image Whitelist which allows kritis-server image.
When deploying  a kritis in test cluster, the server was denied admission as the kritis-server container image was not whitelisted.
( I had a whitelist Image Spec Policy CRD deployed)
We should whitelist kritis-sever image either by shipping a default ImagePolicySpec with the kritis-server image in whitelist.

priyawadhwa:
Add a global image whitelist which the webhook will always admit. This should allow users to update the kritis-server image without having to delete and redeploy the webhook.

Should fix #48


pkg/kritis/util/whitelist_test.go
```diff
@@ -0,0 +1,77 @@
...
+			name: "images in whitelist",
+			images: []string{
+				"gcr.io/kritis-project/kritis-server:tag",
+				"gcr.io/kritis-project/kritis-server@sha256:0000000000000000000000000000000000000000000000000000000000000000",
+			},
+			expected: true,
+		},
+		{
+			name: "some images not whitelisted",
+			images: []string{
```
tejal29: What is the difference in tc 1 and tc 2?

priyawadhwa: the second test case has an image that isn't globally whitelisted, so checks that the function returns false

tejal29: Maybe i am missing something. Arent they the same image names?

priyawadhwa: Oh haha yah I can see how it's not the clearest, the second image is in project kritis instead of kritis-project

tejal29: ah ok. Can you instead add some/image@sha:00000 ?

PR #84, opened 2018-07-18
Fix ISP example spec

priyawadhwa:
The image whitelist isn't meant to be in the spec (oops)


integration/run_test.go
```diff
@@ -178,6 +179,7 @@ func initKritis(t *testing.T) func() {
 		"--set", fmt.Sprintf("serviceNamespace=%s", "default"),
 	)
 	helmCmd.Dir = "../"
+	fmt.Println(helmCmd)
```
dlorenc: println

artifacts/examples/image-security-policy-example.yaml
```diff
@@ -4,12 +4,12 @@ metadata:
...
-  imageWhitelist:
-  - gcr.io/kritis-int-test/nginx-digest-whitelist:latest
-  - gcr.io/kritis-int-test/nginx-digest-whitelist@sha256:56e0af16f4a9d2401d3f55bc8d214d519f070b5317512c87568603f315a8be72
   packageVulnerabilityRequirements:
     maximumSeverity: HIGH
     onlyFixesNotAvailable: true
     whitelistCVEs:
       - providers/goog-vulnz/notes/CVE-2017-1000082
       - providers/goog-vulnz/notes/CVE-2017-1000081
+imageWhitelist:
```
dlorenc: we should actually put this back under spec

priyawadhwa:
nope, i can change it now!

Issue #90, opened 2018-07-19
The labels that we are creating for pods are incorrect.


tejal29:
Fix Background job invalid labels.
```
time="2018-07-19T20:42:08Z" level=error msg="handling violations: Pod \"kritis-validation-hook-54b9f7d9f7-cc5cz\" is invalid: metadata.labels: Invalid value: \"Image not resolved to digest.\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')"
```

dlorenc:
Fixed with #92

PR #92, opened 2018-07-19
Add cron job

### LINKED ISSUE #72: Hook up cron job to main.go
Right now the cron job does not run off when we start the kritis-server.

Add it to main and kick it off.

tejal29:
Fixes #72 and #90

### DIFF
--- a/cmd/kritis/admission/main.go
+++ b/cmd/kritis/admission/main.go
@@ -51,11 +51,7 @@ func main() {
 
 	log.Println("Running the background job in background")
 	// Kick off back ground cron job.
-	e := make(chan error, 1)
-	go StartCronJob(e)
-	if err := <-e; err != nil {
-		log.Fatal(errors.Wrap(err, "Could not start Cron Job due to "))
-	}
+	StartCronJob()
 
 	// Start the Kritis Server.
 	log.Println("Running the server")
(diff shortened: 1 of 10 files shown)


kritis-charts/templates/kritis-server-deployment.yaml
```diff
@@ -23,7 +23,10 @@ spec:
       - name: {{ .Values.image.name }}
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
-        args: ["--tls-cert-file=/var/tls/cert", "--tls-key-file=/var/tls/key"]
+        args: ["--tls-cert-file=/var/tls/cert",
+               "--tls-key-file=/var/tls/key",
+               "--cron-interval=1m",
```
priyawadhwa: Are we actually checking every minute, or is this for integration testing? If it's just for testing we could add the interval to [values-int.yaml](https://github.com/grafeas/kritis/blob/master/integration/values-int.yaml)

cmd/kritis/admission/main.go
```diff
@@ -38,10 +46,20 @@ func main() {
 	flag.StringVar(&tlsCertFile, "tls-cert-file", "/var/tls/tls.crt", "TLS certificate file.")
 	flag.StringVar(&tlsKeyFile, "tls-key-file", "/var/tls/tls.key", "TLS key file.")
 	flag.Set("logtostderr", "true")
+	flag.StringVar(&cronInterval, "cron-interval", "1h", "Cron Job time interval as Duration e.g. 1h, 2s")
 	flag.Parse()
+
+	log.Println("Running the background job in background")
```
dlorenc: nit, this is a bit redundant.

tejal29: will remove

Issue #89, opened 2018-07-19
Fix error "error listing all image policy requirements" when no image policy is configured.


tejal29:
Right now, when we install the Webhook without any image policies we see this error.
```
kubectl create -f exanple.yaml
Error:
 msg="fetching image security policies: error listing all image policy requirements: the server could not find the requested resource (get imagesecuritypolicies.kritis.grafeas.io)"
```
This error disappears when we deploy a image policy spec.
Lets not return false but just return true and log message "No Image policy CRD defined" or ship one by default

PR #94, opened 2018-07-20
Move crd defintions to helm template

priyawadhwa:
I moved the ImageSecurityPolicy and AttestationAuthority CRD definition to templates/ so that they would be created by helm.

I ran into #89 as well because I hadn't created the CRD definition.


tejal29:
This actually wont work.
CRD can be installed once per cluster. They are cluster wide. 
However, objects inside helm templates directory can be installed per namespace.
Hence running helm install kritis-charts/ --namespace=<another namepsace> will give you an error.

I am going to install both the CRDs in the helper script.

tejal29:
dupe of #95
I moved them in templates.  We would want user to install kritis in default namespace unless for testing (like our usecase)

dlorenc:
> I am going to install both the CRDs in the helper script.

Helm has special support for CRDs: https://docs.helm.sh/developing_charts/#defining-a-crd-with-the-crd-install-hook

PR #95, opened 2018-07-20
Add kirits install script ad update Readme.

### LINKED ISSUE #89: Fix error "error listing all image policy requirements" when no image policy is configured.

tejal29:
dupe of #94 
fixes #89


kritis-charts/README.md
```diff
@@ -116,33 +116,12 @@ You will need [helm](https://docs.helm.sh/using_helm/) installed to install krit
...
-   
-   Note: Please install in the same namespace that you created the secret using the `certgen` plugin.
+2. Now run the install script.
+   This should create TLS certs, deploy kritis in given namespace.
+
    ```
-   $ helm install ./kritis-charts --namespace <your namesapce> \
-        --set caBundle=$(kubectl get secret tls-webhook-secret -o jsonpath='{.data.cert}') \
-        --set secret.name=<your secret name>
+    ./install/install-kritis.sh -n <your namespace DEFAUT=default> -s <your secret name | DEFAULT=gac-secret
```
priyawadhwa: nit: <your namespace | DEFAULT=default>

install/install-kritis.sh
```diff
@@ -0,0 +1,70 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+# Command line Args Default
+NAMESPACE="default"
+GAC_SECRET="gac-secret"
```
priyawadhwa: If we're assuming the default secret name is gac-secret, we should change it in [values.yaml](https://github.com/grafeas/kritis/blob/master/kritis-charts/values.yaml#L49) and in the README, where we create secrets.

tejal29: yeah we can do that. Will make those changes.

aaron-prindle:
kokoro didn't run the integration tests here, not sure why

PR #106, opened 2018-07-23
add kritis --version command flag

aaron-prindle:


aaron-prindle:
this also logs the kritis version (git commit) in the kritis-server logs

PR #113, opened 2018-07-23
Int test config

aaron-prindle:


aaron-prindle:
this turns on the integration tests, they were passing when testing locally

integration/run_test.go
```diff
@@ -172,10 +172,11 @@ func initKritis(t *testing.T) func() {
 		t.Fatalf("testing error: %v", err)
 	}
 	helmCmd = exec.Command("helm", "install", "./kritis-charts",
-		"-f", "integration/values-int.yaml",
 		"--namespace", "default",
 		"--set", fmt.Sprintf("caBundle=%s", kubeCA),
-		"--set", fmt.Sprintf("serviceNamespace=%s", "default"),
+		"--set", fmt.Sprintf("image.repository=%s",
+			"gcr.io/kritis-int-test/kritis-server"),
+		"--set", fmt.Sprintf("=%s", "default"),
```
priyawadhwa: is serviceNamespace meant to be set here?

aaron-prindle: it is "placeholder" as at somepoint we might want to switch the namespace.  i changed the values.yaml to make it the default, in the one @ HEAD i don't believe there is a default value set so it is required

PR #124, opened 2018-07-24
Replace helm certgen with preinstall pod which creates CSR and TLS secret

priyawadhwa:
I added a preinstall pod which runs a script to create the CertificateSigningRequest and TLS secret required to start kritis. This way we can remove any dependency on the certgen plugin.

I still need to:
- Fix up the docs 
- Potentially add this to the integration tests
- Deploy the preinstall image via build trigger


preinstall/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+FROM golang:1.10
+
+RUN go get -u github.com/cloudflare/cfssl/cmd/...
+
+ENV KUBECTL_VERSION v1.10.0
+RUN curl -Lo /usr/local/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl  && \
+    chmod +x /usr/local/bin/kubectl
+
+COPY out/preinstall /preinstall
+ENTRYPOINT /preinstall
```
dlorenc: do you need this entrypoint? Can we make it explicit in the job spec?

preinstall/main.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+var (
+	namespace     string
+	csrName       string
+	tlsSecretName string
+)
+
+func init() {
+	flag.StringVar(&namespace, "namespace", "default", "The namespace you want to install kritis in.")
```
dlorenc: You can get this by reading /var/run/secrets/kubernetes.io/serviceaccount/namespace instead.

install/install-kritis.sh
```diff
@@ -38,17 +38,14 @@ while getopts "n:s" opt; do
...
 done
 
-# creates a cert
-function kritis::create_cert {
-    # First Substitute $NAMESPACE in kritis-charts/cert.yaml
-    cat $CERT_TEMPLATE_FILE | sed "s/NAMESPACE/$NAMESPACE/g" > $CERT_FILE
-    CMD="helm certgen generate $CHARTS_DIR --namespace $NAMESPACE"
-    $CMD
+function kritis::preinstall {
+  cat $PREINSTALL_TEMPLATE_FILE | sed "s/NAMESPACE/$NAMESPACE/g" > $PREINSTALL_FILE
```
dlorenc: You can drop this templating by  leaving namespace out all together and passing it to the kubectl apply -f below.

preinstall/preinstall.go
```diff
@@ -0,0 +1,147 @@
...
+	"fmt"
+	"github.com/sirupsen/logrus"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+func deleteExistingObjects() {
+	csrCmd := exec.Command("kubectl", "get", "csr", csrName)
```
tejal29: namespace should go over here too

preinstall/preinstall.go
```diff
@@ -0,0 +1,147 @@
...
+	if err == nil {
+		deleteCSRCmd := exec.Command("kubectl", "delete", "csr", csrName)
+		RunCommand(deleteCSRCmd)
+	}
+	secretCmd := exec.Command("kubectl", "get", "secret", tlsSecretName)
+	secretCmd.Stderr = os.Stderr
+	_, err = secretCmd.Output()
+	fmt.Println(err)
+	if err == nil {
+		deleteSecretCmd := exec.Command("kubectl", "delete", "secret", tlsSecretName)
```
tejal29: add namespace.

preinstall/preinstall.go
```diff
@@ -0,0 +1,147 @@
...
+	serverCmd.Stdin = bytes.NewReader(output)
+	RunCommand(serverCmd)
+}
+
+func createCertificateSigningRequest() {
+	certificate := retrieveRequestCertificate()
+	csr := `apiVersion: certificates.k8s.io/v1beta1
+kind: CertificateSigningRequest
+metadata:
+    name: %s
```
tejal29: i think namespace shd go over here.

preinstall/preinstall.go
```diff
@@ -0,0 +1,147 @@
...
+	// base64 encode the contents
+	encodedLen := base64.StdEncoding.EncodedLen(len(contents))
+	encoded := make([]byte, encodedLen)
+	base64.StdEncoding.Encode(encoded, contents)
+	// trim any new lines off the end
+	return string(encoded)
+}
+
+func createTLSSecret() {
+	retrieveCertCmd := exec.Command("kubectl", "get", "csr", csrName, "-o", "jsonpath='{.status.certificate}'")
```
tejal29: +namespace.

preinstall/preinstall.go
priyawadhwa: The CertificateSigningRequests are global rather than namespaced

kritis-charts/templates/kritis-server-deployment.yaml
```diff
@@ -23,8 +23,8 @@ spec:
       - name: {{ .Values.image.name }}
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
-        args: ["--tls-cert-file=/var/tls/cert",
-               "--tls-key-file=/var/tls/key",
+        args: ["--tls-cert-file=/var/tls/tls.crt",
```
tejal29: The secret will have data fields "cert" and "key" since you are using kubectl create tls --cert --key command.

PR #137, opened 2018-07-26
[Do not Review yet] Add cron job demo

### LINKED ISSUE #120: Add steps to run and Demo Cron Job.
1. Add steps to run the cron job from binary.
2. Demo the flow where you 
   a. Deploy a pod with vulnerable image with breakglass annotation  
   b. Remove the annotation with kubectl edit command
   c. Start the cron job with 1 min param
   d. You shd see the annotation added to pod.

tejal29:
Fixes #120

This is still WIP but it out there for initial feedback.

TODO:
- Add unit test if applicable
- Add Integration test.

### DIFF
--- a/pkg/kritis/admission/admission.go
+++ b/pkg/kritis/admission/admission.go
@@ -22,11 +22,14 @@ import (
 	"io/ioutil"
 	"net/http"
 
+	"github.com/grafeas/kritis/pkg/kritis/secrets"
+
 	"github.com/golang/glog"
 	"github.com/grafeas/kritis/cmd/kritis/version"
 	"github.com/grafeas/kritis/pkg/kritis/admission/constants"
 	kritisv1beta1 "github.com/grafeas/kritis/pkg/kritis/apis/kritis/v1beta1"
 	kritisconstants "github.com/grafeas/kritis/pkg/kritis/constants"
+	"github.com/grafeas/kritis/pkg/kritis/container"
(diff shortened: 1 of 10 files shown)


pkg/kritis/admission/admission.go
```diff
@@ -61,13 +61,25 @@ var (
...
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
+
+	// Check, if the pod is already running. This handles scenarios when security
+	// policy is deployed after a pod is deployed.
+	// In this situation, the cron job will annotate the pod and call v1.CoreV1.Pod.Patch()
+	// kubectl patch command goes through admission review before patching the pod and will return error.
+	// In order to apply the labels for invalid pods, we skip admission review for already running pods.
+	if pods.IsPodRunning(*pod) {
```
dlorenc: Does the cron job come through this function (AdmissionReviewHandler)?

tejal29: the cron job does not come through this function. But it call `kubectl apply patch` for adding annotations which go through admission review.
So, the labels never get applied.

tejal29: Now, that i think more about it, the admission controller will always return "true" on updates.
Kind of defeats the purpose of having this [here](https://github.com/grafeas/kritis/blob/master/kritis-charts/templates/kritis-server-webhook.yaml#L14) 

I think, this will be simpler and better once we add attestations to indicate if this pod was ever admitted.

PR #143, opened 2018-07-27
Update templates to allow installing kritis in a different namespace

priyawadhwa:
Updated templates and the install README


aaron-prindle:
to allow for integration tests to be namespaced we also need to reuse the csr somehow

aaron-prindle:
or disable the pre-install hooks

priyawadhwa:
I added a `--delete-csr` flag in #130, we can set it to false in the integration tests so that the same CSR is used every time.

priyawadhwa:
Oh, but I think we'll need to add the flag to the preinstall pod template for it to work!

PR #149, opened 2018-07-31
Fix global whitelist and add integration tests

### LINKED ISSUE #142: Check image in Whitelist per image.
I just realized we have a have bug in the way we do the whitelisting check.
Right now, 
1. For a given pod, we check if all images are whitelisted.
  -  If all are, then we return true. 
 - If not, we continue with vuln check.

With this logic, we will return false for a pod which has multiple containers with say 2 images. One whitelisted (but has vuln) and second one not in whitelist but meets the policy.
Since we do the check for all images at the start, CheckGlobalWhitelist will return false.
We will fetch the vuln info for the whitelisted image and return admission response false.

priyawadhwa:
Should fix #142 

I also created `make integration-local` to build new versions of each image so integration tests can be run locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,7 @@ GO_LDFLAGS += -X github.com/grafeas/kritis/cmd/kritis/version.Version=$(VERSION)
 GO_LDFLAGS += -w -s # Drop debugging symbols.
 
 REGISTRY?=gcr.io/kritis-project
+TEST_REGISTRY?=gcr.io/kritis-int-test
 REPOPATH ?= $(ORG)/$(PROJECT)
 SERVICE_PACKAGE = $(REPOPATH)/cmd/kritis/admission
 KRITIS_PROJECT = $(REPOPATH)/kritis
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -115,13 +112,24 @@ clean:
...
+	go test -ldflags "$(GO_LDFLAGS)" -v -tags integration $(REPOPATH)/integration -timeout 5m -- --remote=true
+
 .PHONY: build-push-image
 build-push-image: build-image preinstall-image postinstall-image predelete-image
 	docker push $(REGISTRY)/kritis-server:$(IMAGE_TAG)
 	docker push $(REGISTRY)/preinstall:$(IMAGE_TAG)
 	docker push $(REGISTRY)/postinstall:$(IMAGE_TAG)
 	docker push $(REGISTRY)/predelete:$(IMAGE_TAG)
 
+.PHONY: build-push-test-image
```
aaron-prindle: nice!

PR #151, opened 2018-08-01
Update cloudbuild to release all images

priyawadhwa:


aaron-prindle:
the error here is unrelated, LGTM

PR #153, opened 2018-08-01
Add troubleshooting section

### LINKED ISSUE #104: Add troubleshooting section
* How to get logs
* How to remove the `ValidatingWebhookConfiguration` if kritis is down or failing

### LINKED ISSUE #118: Add a trouble shooting section with all RBAC errors.
RBAC errors seen:
 - helm could not make a release.

priyawadhwa:
Added troubleshooting section and fixed a bug in the Dockerfile which was causing the build trigger to fail

Should fix #118


tejal29:
We need to add help on errors mentioned in the comments #118

priyawadhwa:
> We need to add help on errors mentioned in the comments #118

I think the first error doesn't apply since we aren't using the plugin anymore, and the third error fix is in the install README already (in the Installing Helm section, sorry I'm not sure how to link to it!)

For the second error, the only thing I think we could say is to make sure you run under an account that has permissions to create service accounts. I can add that in!

Issue #148, opened 2018-07-31
Change tutorials to push image on user personal project.


tejal29:
Right now, there is no way, user can access vulnerability data on public images.
Users will have to use images pushed to their project.

Change the Tutorial to do so:
- [x] Create a project, where user want to push their images/
- [x] Enable Container analysis Api and Vulnerability Scanning.
- [ ] add optional steps to make the images public or a note ( for tutorial we dont need the image bucket to be public)
- [x] Add a script which user will have to run to push all the test images into their project
- [x] Remove all instances of GCP project "kritis-int-test" and replace them with user project.

tstromberg:
It's noteworthy that the API's mentioned are in Alpha, and that users have to apply to be white-listed for the Container Analysis API and Vulnerability Scanning API. Instructions are here:

https://cloud.google.com/container-registry/docs/get-image-vulnerabilities

tejal29:
yes. we are using alpha libs. We have an issue here #27. We are waiting for the libs.

tstromberg:
Working now on understanding the image copy

tstromberg:
Fixed by #197

It does not cover the optional "make images public" step, because it doesn't seem useful for the tutorial as it stands.

PR #155, opened 2018-08-02
Make it easier to find the installation and tutorial

tstromberg:
This moves the installation guide up one directory, and splits the tutorial out into its own file.

This is in preparation for ensuring that the guides are both comprehensive and transition cleanly from one another. See #148 for more context.


tejal29:
Looks Really great!

PR #112, opened 2018-07-23
Add attestation logic PR 1 of 4.

tejal29:
1.  In this PR just create notes from Attestation Authority.

Next PRs:
1. get public and private keys from kubernetes secret (Coming soon)
2. Add attestation Occurrences (Coming Soon)
3. Finally add attestation to flow.

PR #157, opened 2018-08-02
Refactor and fix gpg signing Verification test.

tejal29:
Refactoring changes
1. Move AtomicContainerSig from util to its own package container
2. Add pkg/kritis/util/metadata_mock.go which defines metadata mock used in admission_test.go and securitypolicy_test.go
3. Add functions to MetadataFethcer Interface which i added in Pr #112 and update the mock to return those.

Fixes in Attestation Lib.
So after debugging the attestation_lib, i found out that that openpgp.ReadMessage returns an error and openpgp.MessadeDetails  when reading a signed message.
The lib relies on two fields, SignatureError and Signature if a valid signature for the given public is found or not, 
Here are the rules `openpgp.ReadMessage(signatureblock reader, keys, <prompt func for accepting passwd>, pgpConfig) (md *MessageDetails, err error)`
If, for a given public key, 
- Invalid Input, i.e. Signature and or keys can not be parsed: Return  <nil, error>.

- Signature not Present for a public key loaded in keys: Return  <md, nil>.
    - md.Signature is  nil and 
    - md.SinatureError is also nil since it never executed the signature verification. 
   Note: This does not return an error as there was no error reading the signed message.

- Signature Present for a public key loaded in keys but could not be verified: Return  <md, nil>
   - md.UnverifiedBody will contain the [reader](https://github.com/golang/crypto/blob/master/openpgp/read.go#L321) which verifes signature using PublicKey.VerifySignature() method.
  Now, in order to verify the signature, you need to read md.UnverifiedBody.
  - md.Signature is not nil and contains the signature.
  - md.SignatureError contains the error indicating why the signature could not be verified.

- Signature Present for a public key loaded in keys and can be verified: Return <md, nil>
Similar to case above, 
  - md.UnverifiedBody will contain the [reader](https://github.com/golang/crypto/blob/master/openpgp/read.go#L321) which verifes signature using PublicKey.VerifySignature() method.
  Now, in order to verify the signature, you need to read md.UnverifiedBody.
  - md.Signature is not nil and contains the signature.
  - md.SignatureError does not contain error.
This means the signature was valid for the public key(s) given as input to this message. However, this signature could be a valid signature over any message. So now we need to make sure the plaintext which we got after reading the md.UnverifiedBody is same as message we want to verify.

Let me know if all this does not make sense.

Thanks
Tejal


pkg/kritis/metadata/containeranalysis/containeranalysis.go
```diff
@@ -72,8 +72,16 @@ func (c ContainerAnalysis) GetVulnerabilities(containerImage string) ([]metadata
 }
 
 // GetAttestation gets AttesationAuthority Occurrences for a specified image.
-func (c ContainerAnalysis) GetAttestations(containerImage string) ([]*containeranalysispb.Occurrence, error) {
-	return c.fethcOccurrence(containerImage, AttestationAuthority)
+func (c ContainerAnalysis) GetAttestations(containerImage string) ([]metadata.PgpAttestation, error) {
+	occs, err := c.fethcOccurrence(containerImage, AttestationAuthority)
```
priyawadhwa: nit: fetchOccurrence

pkg/kritis/admission/admission_test.go
```diff
@@ -78,12 +78,10 @@ func Test_UnqualifiedImage(t *testing.T) {
...
 		return []kritisv1beta1.ImageSecurityPolicy{{}}, nil
 	}
-	mockMetadata := func() (metadata.MetadataFetcher, error) {
-		return mockMetadataClient{}, nil
-	}
+
 	mockConfig := config{
 		retrievePod:                 mockPod,
-		fetchMetadataClient:         mockMetadata,
+		fetchMetadataClient:         testutil.EmptyMockMetadata(),
```
tstromberg: A few testing readability comments here:

- I'd rather see the mock metadata inline than moved into a helper method. When writing tests, one should not sacrifice clarity in the name of de-duplication.
- "Empty" is a misnomer, as the helper does in fact setup attestation and severity.
- I can't tell if it's the case here, but tests should never rely on default values specified by a helper method, since that forces readers to read the helper method’s implementation details in order to understand the test.

pkg/kritis/attestation/attestation.go
```diff
@@ -53,27 +52,40 @@ func VerifyMessageAttestation(pubKeyEnc string, attestationHash string, message
...
+	}
+	// Make sure after reading the UnverifiedBody above, there is no signature error.
+	if md.SignatureError != nil || md.Signature == nil {
+		return fmt.Errorf("Bad signature found: %s or No signature found for given key", md.SignatureError)
 	}
 
+	// Finally, make sure the signature is over the right message.
 	if string(plaintext) != message {
-		return fmt.Errorf("Signature could not be verified. Got: %q, Want: %q", plaintext, message)
+		return fmt.Errorf("Signature could not be verified. Got: %s, Want: %s", string(plaintext), message)
```
tstromberg: https://github.com/golang/go/wiki/Errors

pkg/kritis/attestation/attestation_test.go
```diff
@@ -50,6 +51,30 @@ func TestAttestations(t *testing.T) {
 	}
 }
 
+func TestGPGArmorSignIntegration(t *testing.T) {
+	if err := VerifyMessageAttestation(testutil.PublicTestKey, base64.StdEncoding.EncodeToString([]byte(expectedSig)), "test"); err != nil {
+		t.Fatalf("unexpected error %s", err)
+	}
+}
+
+// Base64 encoded signarute.
```
tstromberg: spelling

pkg/kritis/attestation/attestation_test.go
```diff
@@ -50,6 +51,30 @@ func TestAttestations(t *testing.T) {
...
+// Created using gpg --armor --sign -u test@kritis.org < test | base64
```
tstromberg: what are the contents of test?

pkg/kritis/metadata/containeranalysis/containeranalysis.go
```diff
@@ -72,11 +72,19 @@ func (c ContainerAnalysis) GetVulnerabilities(containerImage string) ([]metadata
...
 
 // GetAttestation gets AttesationAuthority Occurrences for a specified image.
-func (c ContainerAnalysis) GetAttestations(containerImage string) ([]*containeranalysispb.Occurrence, error) {
-	return c.fethcOccurrence(containerImage, AttestationAuthority)
+func (c ContainerAnalysis) GetAttestations(containerImage string) ([]metadata.PgpAttestation, error) {
+	occs, err := c.fetchOccurrence(containerImage, AttestationAuthority)
+	if err != nil {
+		return nil, err
+	}
+	pgpAttestations := make([]metadata.PgpAttestation, len(occs))
```
tstromberg: Variable names in Go should be short rather than long. This is especially true for local variables with limited scope. Prefer c to lineCount. Prefer i to sliceIndex.

Maybe just "p" here? The variable lifetime is very short, and it's already clear from the context that this is a slice of PgpAttestations.

pkg/kritis/metadata/metadata.go
```diff
@@ -16,13 +16,32 @@ limitations under the License.
...
+	GetAttestations(containerImage string) ([]PgpAttestation, error)
 }
 
 type Vulnerability struct {
 	Severity        string
 	HasFixAvailable bool
 	CVE             string
 }
+
+type PgpAttestation struct {
```
tstromberg: PGPAttestation? Words in names that are initialisms or acronyms (e.g. "URL" or "NATO") have a consistent case. For example, "URL" should appear as "URL" or "url" (as in "urlPony", or "URLPony"), never as "Url".

tstromberg: Also, since this is a public struct, should it get documentation?

pkg/kritis/testutil/util.go
```diff
@@ -84,3 +85,12 @@ func getKey(key *openpgp.Entity, keyType string, t *testing.T) []byte {
 	wr.Close()
 	return gotWriter.Bytes()
 }
+
+func CreateSecret(t *testing.T, name string) *secrets.PgpSigningSecret {
```
tstromberg: This helper seems trivial enough that it's not worth the additional cognitive overhead of needing to read a second file to understand the test.

pkg/kritis/attestation/attestation.go
tejal29: thanks! will fix!

pkg/kritis/metadata/metadata.go
tejal29: we haven't documented most of the public structs. I will add one for this one and probably we can go back add later for other structs.

pkg/kritis/metadata/containeranalysis/containeranalysis.go
tejal29: thanks. The reason why i had "Pgp" Vs "PGP" was because the containeranalysis proto has it as PgpSignedAttestation.
I can change it to fix it.

pkg/kritis/testutil/util.go
tejal29: okies. i moved it to the file where we are using it. In case we need it else where, we can move it to test util.

pkg/kritis/admission/admission_test.go
tejal29: hmm, the reason why i moved it to helper is because, you need to do a bunch of mock setups, instead of focusing on the values you want to test your code against.
The idea of these mocks is users can just pass the vulnerabilites and attestations that image has.
say, i want to write a test case where my-isp allows max severity LOW and the image has a HIGH severity vuln, i can use the helper to do so, 
```
vulnzHigh = metadata.Vulnerability{
		CVE:             "cve2",
		Severity:        "HIGH",
		HasFixAvailable: false,
	}
```
And setup the mock to pass this value using.
```
mockMetadataClient = testutil.MockMetadataClient{
		Vulnz: []metadata.Vulnerability{
			vulnz1,
		},
		PGPAttestations: nil,
	}
// test my code.
```

VS create a MetadataFetcher interface in you tests and implementing all the required methods:
```
type mockMetadataClient struct {
}

func (m mockMetadataClient) GetVulnerabilities(containerImage string) ([]metadata.Vulnerability, error) {
	return []metadata.Vulnerability{
		vulnHigh,
	}, nil
}

func (m mockMetadataClient) CreateAttestationOccurence(note *containeranalysispb.Note,
	containerImage string,
	pgpSigningKey *secrets.PGPSigningSecret) (*containeranalysispb.Occurrence, error) {
	return nil, nil
}

func (m mockMetadataClient) GetAttestationNote(aa kritisv1beta1.AttestationAuthority) (*containeranalysispb.Note, error) {
	return nil, nil
}

func (m mockMetadataClient) GetAttestations(containerImage string) ([]metadata.PGPAttestation, error) {
	return <some attestations>, nil
}

// test my code.
```
Doing this for every test case, would be too cumbersome. You can declare `mockMetadataClient`  struct which accepts vuln and pgp attestations and return those. You will have to repeat that code in every test file. The helper just pulls it out and provides a public interface which you can use everywhere.

Does that make sense?

Regarding EmptyMockMetadata: EmptyMockMetadata() returns no vulnerabilities and attestations. Maybe i shd i rename it to `NilMetadataFethcer.`

pkg/kritis/attestation/attestation.go
tstromberg: Also, why call string(plaintext) - won't fmt.Errorf do that for you?

tejal29: yes it does. copy paste issue.

pkg/kritis/admission/admission_test.go
tstromberg: Seems reasonable.

In tests, repeated code is far more preferable to de-duplicated code, but there is a healthy balance to be made. Tests should be optimized for readability, not writability, since readers will always approach tests with an air of skepticism, and authors will approach them with an air of confidence.

pkg/kritis/attestation/pgpkey.go
```diff
@@ -96,7 +90,11 @@ func parsePrivateKey(privateKey string) (*packet.PrivateKey, error) {
 }
 
 func parseKey(key string, keytype string) (packet.Packet, error) {
-	r := strings.NewReader(key)
+	pemKeyString, err := base64.StdEncoding.DecodeString(key)
```
tstromberg: When naming variables, omit words that are obvious given a variable's type declaration.

In Go, it's idiomatic to prefer short names for such a short-lived variable, like "s".

pkg/kritis/metadata/containeranalysis/containeranalysis.go
tstromberg: Ahh, it's a proto. Proto's always have braindead capitalization :(

PR #165, opened 2018-08-02
Run integration tests in unique namespaces

### LINKED ISSUE #163: allow for multiple kritis instances on a single cluster (parallelize int test)
currently users can only run one instance of kritis on a cluster at once.  This is useful for users upgrading and running kritis and it will also allow integration tests to be parallelized.  Currently the following blockers remain:

- [ ] per instance webhook CSR, per instance tls Secret

priyawadhwa:
Allows us to run integration tests in parallel in different namespaces.

Depends on #154  to merge first.

Should fix #163


hack/kokoro-test.sh
```diff
@@ -28,6 +28,8 @@ pushd $GOPATH/src/github.com/grafeas/kritis
 echo "Check format"
 ./hack/check-fmt.sh
 
+echo "Running gcloud auth login..."
+gcloud auth login
```
tstromberg: What's this for? It's not mentioned in the pull description.

integration/run_test.go
```diff
@@ -155,20 +126,27 @@ func createCRDExamples(t *testing.T) {
...
-func deleteFailedDeployments() {
-	deleteWebhookCmd := exec.Command("kubectl", "delete", "validatingwebhookconfiguration", "--all")
-	integration_util.RunCmdOut(deleteWebhookCmd)
-	helmCmd := exec.Command("sh", "-c",
-		"helm ls --short | xargs -L1 helm delete")
-	helmCmd.Dir = "../artifacts"
-	integration_util.RunCmdOut(helmCmd)
-}
+func createGACSecret(t *testing.T, ns *v1.Namespace) {
+	crdCmd := exec.Command("gsutil", "cp", "gs://kritis-test-files/gac.json", "/tmp")
```
tstromberg: Ignore if this is impossible to handle otherwise, but I worry that this might be causing the tests to more non-hermetic than necessary, by introducing a dependency on an external data source that might change outside of the test.

Is it possible to generate the secret on the fly?

integration/run_test.go
```diff
@@ -448,22 +430,19 @@ func TestKritisPods(t *testing.T) {
...
+	deleteKritis := initKritis(t, ns)
 	defer deleteKritis()
-	if err := kubernetesutil.WaitForDeploymentToStabilize(client, "default",
+	if err := kubernetesutil.WaitForDeploymentToStabilize(client, ns.Name,
 		"kritis-validation-hook", 2*time.Minute); err != nil {
 		t.Fatalf("Timed out waiting for deployment to stabilize")
 	}
 	createCRDExamples(t)
-	time.Sleep(5 * time.Second)
+	time.Sleep(10 * time.Second)
```
tstromberg: Why double the sleep? Is it possible to replace the sleep statement with something that polls for when the time is appropriate to proceed?

hack/kokoro-test.sh
priyawadhwa: I'm trying to figure out auth in kokoro for reading from the kritis-int-test GCS bucket. It gave me an error "run gcloud auth login" so I figured I'd try it, but doesn't look like it worked :(

Will remove & try to find another way!

integration/run_test.go
priyawadhwa: since the tests run in a different namespace every time we need to generate the secret during runtime, but I switched this from copying the creds from a GCS bucket to copying them directly into kokoro first.

integration/run_test.go
priyawadhwa: For sure, I ended up moving createCRDExamples() before the check for the stabilized deployment which should (hopefully) be enough time

integration/run_test.go
```diff
@@ -38,14 +38,20 @@ import (
...
+const (
+	kritisPreinstall  = "kritis-preinstall"
+	kritisPostinstall = "kritis-postinstall"
+)
 
-var context *api.Context
+var (
+	gkeZone        = flag.String("gke-zone", "us-central1-a", "gke zone")
+	gkeClusterName = flag.String("gke-cluster-name", "kritis", "name of the integration test cluster")
+	gcpProject     = flag.String("gcp-project", "priya-wadhwa", "the gcp project where the integration test cluster lives")
```
tstromberg: The default value here seems unusual. Was this unintentional?

priyawadhwa: Whoops forgot to change that back, thanks :sweat_smile:

integration/run_test.go
aaron-prindle: You should be able to check if the ISP is up by looking waiting for a successful return code for 'kubectl get imagesecuritypolicy my-isp'.  The stabilized deployment check won't really help verify that the ISP is created

priyawadhwa: oh perfect, i'll add that in, thanks!

Issue #160, opened 2018-08-02
Verify tutorial


aaron-prindle:
Verify Tutorial and add steps to push images to user project before running kritis

tstromberg:
Image pushing is a dupe of #148

tstromberg:
Tutorial now has steps to copy images into ones project. Tutorial has been run through a couple of times, and save for resolve-tags issues, seems to work.

PR #169, opened 2018-08-03
First pass at making install guide repeatable and easier to follow

tstromberg:
Includes all the API enabling steps required, fixes typos, and applies some (but not all) of the recommondations from the https://kubernetes.io/docs/home/contribute/style-guide/

Works toward addressing #160 and #148.

### DIFF
--- a/install.md
+++ b/install.md
@@ -1,8 +1,18 @@
 # Installing Kritis
 
+## Requirements
+
+The only supported backend for vulnerability data is the [Google Cloud Container Analysis API](https://cloud.google.com/container-registry/docs/container-analysis). The current requirements for installation are:
+
+- [Google Cloud](https://cloud.google.com) account with [billing](https://console.cloud.google.com/billing) enabled
+- [Kubernetes](https://kubernetes.io/) 1.9.2+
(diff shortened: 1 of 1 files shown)


install.md
```diff
@@ -1,240 +1,227 @@
...
-1. First create the service account
-```
-gcloud iam service-accounts create $ACC_NAME --display-name "$ACC_DISP_NAME"
-```
-This should create a service account $ACC_NAME@$PROJECT_ID.iam.gserviceaccount.com.
+For more documentation, see [Container Analysis Overview](https://cloud.google.com/container-registry/docs/container-analysis). 
+
+## Step #4: Create a cluster
+
+You may skip this step if you already have a cluster configured.
```
tejal29: the cluster needs to be version  > 1.9.2. Add  note for that.
Also, we had the version number mentioned on the `gcloud cluster create`. Did you drop that by mistake or the default is now 1.9.3?

tstromberg: Added a note. The default is now 1.9.7: https://cloud.google.com/kubernetes-engine/versioning-and-upgrades

PR #172, opened 2018-08-03
Another refactor to make sure we are using the same function to validate policies in cron job and admission controller

tejal29:
We need to attest pods in Cron job as well as Admission controller.
I realized we do not use the same function to review policies in cron and admission controller. 
I was doing the same thing twice and hence the refactor.
This refactor, 
-  Move the review function from admission.reviewImages to its own package.
- Added a todo to move the tests pertaining to review.Review to review_test.go.
- Both admission and cron job use the same function.

### DIFF
--- a/pkg/kritis/admission/admission.go
+++ b/pkg/kritis/admission/admission.go
@@ -31,7 +31,7 @@ import (
 	"github.com/grafeas/kritis/pkg/kritis/metadata"
 	"github.com/grafeas/kritis/pkg/kritis/metadata/containeranalysis"
 	"github.com/grafeas/kritis/pkg/kritis/pods"
-	"github.com/grafeas/kritis/pkg/kritis/util"
+	"github.com/grafeas/kritis/pkg/kritis/review"
 	"github.com/grafeas/kritis/pkg/kritis/violation"
 	"k8s.io/api/admission/v1beta1"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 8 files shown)


pkg/kritis/admission/admission.go
```diff
@@ -42,24 +42,22 @@ import (
 )
 
 type config struct {
-	retrievePod                 func(r *http.Request) (*v1.Pod, v1beta1.AdmissionReview, error)
-	retrieveDeployment          func(r *http.Request) (*appsv1.Deployment, v1beta1.AdmissionReview, error)
-	fetchMetadataClient         func() (metadata.MetadataFetcher, error)
-	fetchImageSecurityPolicies  func(namespace string) ([]kritisv1beta1.ImageSecurityPolicy, error)
-	validateImageSecurityPolicy func(isp kritisv1beta1.ImageSecurityPolicy, image string, client metadata.MetadataFetcher) ([]securitypolicy.SecurityPolicyViolation, error)
```
tejal29: validateImageSecurityPolicy was not used anywhere in the tests.

pkg/kritis/review/review.go
```diff
@@ -0,0 +1,77 @@
...
+
+func New(client metadata.MetadataFetcher, vs violation.Strategy, validate securitypolicy.ValidateFunc) Reviewer {
+	return Reviewer{
+		client:   client,
+		vs:       vs,
+		validate: validate,
+	}
+}
+
+// Review reviews a given images against ImageSecurityPolicies and return error
```
tstromberg: Some grammar/spelling issues here.

How about just: "Review a set of images against a set of policies"?

pkg/kritis/review/review.go
```diff
@@ -0,0 +1,77 @@
...
+			if len(violations) != 0 {
+				errMsg := fmt.Sprintf("found violations in %s", image)
+				// Check if one of the violations is that the image is not fully qualified
+				for _, v := range violations {
+					if v.Violation == securitypolicy.UnqualifiedImageViolation {
+						errMsg = fmt.Sprintf("%s is not a fully qualified image", image)
+					}
+				}
+				if err := r.vs.HandleViolation(image, pod, violations); err != nil {
+					return fmt.Errorf("%s. error handling voilation %v", errMsg, err)
```
tstromberg: spelling: violation

pkg/kritis/cron/cron.go
```diff
@@ -91,22 +89,16 @@ func Start(ctx context.Context, cfg Config, checkInterval time.Duration) {
...
-					return err
-				}
-				if len(v) != 0 {
-					if err := cfg.ViolationStrategy.HandleViolation(c, &p, v); err != nil {
-						glog.Errorf("handling violations: %s", err)
-					}
-				}
+			glog.Infof("Checking po %s", p.Name)
+			if err := r.Review(pods.Images(p), isps, &p); err != nil {
+				glog.Error(err)
```
tstromberg: This doesn't return an error: how does the CheckPods() caller know there was a violation?

tejal29: CheckPods returns an error if there was error checking a pod. It does not care about violations if found.  The violationStrategy is supposed to handle it. 
Does that make sense?

tstromberg: Makes sense, thanks.

PR #180, opened 2018-08-06
Delete webhooks/csr if helm delete fails in integration tests

### LINKED ISSUE #174: Delete leftover webhooks before running integration tests

priyawadhwa:
Should fix #174

### DIFF
--- a/integration/cleanup.go
+++ b/integration/cleanup.go
@@ -1,0 +1,36 @@
+/*
+Copyright 2018 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


tejal29:
Can we instead just delete the namespace?

Regarding csr, we shd think about parameterizing a  csr value.

priyawadhwa:
The CSR isn't namespaced, so we have to do it separately (it's meant to be deleted in the predelete pod, but this is in case that fails we can delete it manually)

We do parametrize the CSR names in that there is one per namespace that is created for the tests. 

Mostly this is so that the cluster doesn't have a bunch of stale CSRs lying around.

PR #181, opened 2018-08-06
Catch invalid thresholds and make vulnerabilities programmable by the test.

tstromberg:
This is to prepare for making adjustments to how OnlyFixesNotAvailable are handled,
but in general the adjustments are influenced by the following guidelines:

- The test should be so simple as to be correct by inspection
- Readers should need to read only the test body to understand the test
- Tests should never rely on default values specified by a helper method

severityWithinThreshold() now takes two strings instead of an isp and a
string, to both reduce unnecessary coupling (Law of Demeter), and simplify test setup.


aaron-prindle:
thanks for updating these!

Issue #135, opened 2018-07-25
add breakglass to deployment


aaron-prindle:
add breakglass annotation to deployments as well

tejal29:
is this done? @priyawadhwa were you looking at this?

priyawadhwa:
I think @aaron-prindle mentioned it was on his agenda in the email sync?

aaron-prindle:
I am working on this, should be in soon

PR #184, opened 2018-08-07
Breakglass deployments

aaron-prindle:


pkg/kritis/admission/admission.go
```diff
@@ -243,6 +247,15 @@ func checkBreakglass(pod *v1.Pod) bool {
 	return ok
 }
 
+func checkBreakglassDeployment(deployment *appsv1.Deployment) bool {
```
priyawadhwa: Not sure if this would work, but could we have something like

```golang
func checkBreakglass(meta *ObjectMeta){}
```

so that we don't have to separate the checks for pods and deployments? Just wondering if that could work so we can use it for future types we add as well.

integration/run_test.go
```diff
@@ -336,6 +336,27 @@ func TestKritisPods(t *testing.T) {
...
+			},
+			shouldSucceed: true,
+			dir:           "../",
+			cleanup: func(t *testing.T) {
+				cmd := exec.Command("kubectl", "delete", "-f",
+					"integration/testdata/java/java-with-vuln-breakglass-deployment.yaml")
+				cmd.Dir = "../"
+				output, err := integration_util.RunCmdOut(cmd)
+				if err != nil {
+					t.Fatalf("kritis: %s %v", output, err)
```
tstromberg: Two comments: "kritis" doesn't seem like a very informative string here, when the error is relating to deletion. 

Also, is t.Fatalf the right choice? Is it possible for the test continue onward so that we don't have to play whack-a-mole with test failures?

pkg/kritis/admission/admission.go
```diff
@@ -152,6 +152,10 @@ func AdmissionReviewHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func reviewDeployment(deployment *appsv1.Deployment, ar *v1beta1.AdmissionReview) {
+	if checkBreakglassDeployment(deployment) {
+		glog.Infof("found breakglass annotation, returning successful status")
```
tstromberg: This log message could be more useful if it mentioned the name of the deployment it found an annotation for.

integration/run_test.go
aaron-prindle: added more verbose logs to where "kritis:" used to be.

regarding t.Fatalf, the entire set of tests runs each time regardless of single failure, the t.Fatalf only stops the single test in the scope of the t.Run.

pkg/kritis/admission/admission.go
aaron-prindle: wasn't able to get this to work, keeping for now

pkg/kritis/admission/admission.go
aaron-prindle: added names of deployment and pod for these error messages

aaron-prindle:
fixes #135

PR #185, opened 2018-08-07
Refactor img tag

aaron-prindle:


kritis-charts/values.yaml
```diff
@@ -5,8 +5,9 @@
 replicaCount: 1
 
 image:
-  repository: gcr.io/kritis-project/kritis-server
+  repository: kritis-server
```
priyawadhwa: nit: since this field is called "image" for the preinstall/postinstall/predelete images, could you pick one and make it consistent? thanks!

PR #187, opened 2018-08-07
Add linter scripts from Skaffold.

tstromberg:
This won't be enabled as a presubmit test until the underlying lint
issues are fixed.


aaron-prindle:
we have a boilerplate check for certain file types in the repo, including .sh.  I believe you will need to add:
https://github.com/grafeas/kritis/blob/master/hack/boilerplate/boilerplate.sh.txt

to the top of the files

aaron-prindle:
I think you'll also have to add the script itself to make test to run on each commit:
https://github.com/grafeas/kritis/blob/master/Makefile#L43

PR #197, opened 2018-08-08
Improve tiller RBAC configuration and remove kritis-int-test from tutorial.

tstromberg:
Also includes a healthy amount of de-linting, moving tutorial bits around, and addressing doc style concerns.


tutorial.md
```diff
@@ -17,80 +30,83 @@ metadata:
...
 ```
 
-The `ImageSecurityPolicy` should be configured:
+### 2. Upload a vulnerable image
+
+The [Container Analysis API](https://cloud.google.com/container-analysis/api/reference/rest/) only reveals vulnerability information for images owned by your project. This makes a copy of a sample vulnerable image into your container registry:
+
 ```shell
-imagesecuritypolicy.kritis.grafeas.io/my-isp configured
+% gcloud container images add-tag \
```
priyawadhwa: is the % meant to be here?

tutorial.md
```diff
@@ -17,80 +30,83 @@ metadata:
...
 
-### 2. Deploying An Image With Vulnerabilities
-Now that we have kritis installed and an `ImageSecurityPolicy` to validate against, we can go ahead and start deploying some images!
+Deploy a pod containing our vulnerable image:
 
-First, let's try to deploy a java image:
 ```shell
 cat <<EOF | kubectl apply -f - \
-
 apiVersion: v1
```
priyawadhwa: nit: I think the extra line at Line 45 here might be necessary for copy/pasting directly into a shell, otherwise I get this error
```
$ cat <<EOF | kubectl apply -f - \
apiVersion: v1
kind: Pod
metadata:
  name: java-with-vuln
spec:
  containers:
  - name: java-with-vuln
    image: gcr.io/$PROJECT/java-with-vuln@sha256:b3f3eccfd27c9864312af3796067e7db28007a1566e1e042c5862eed3ff1b2c8
    ports:
    - containerPort: 80
EOF
error: Unexpected args: [apiVersion: v1]
See 'kubectl apply -h' for help and examples.
```

tutorial.md
tstromberg: Not at all. Thanks for noticing!

tutorial.md
tstromberg: Fixed.

tutorial.md
```diff
@@ -16,30 +29,29 @@ metadata:
...
+  gcr.io/$PROJECT/java-with-vuln:latest
 ```
-This `ImageSecurityPolicy` specifies two `nginx` images that are whitelisted and should always be allowed entry.
-It sets the maximum CVE severity allowed in any image to `HIGH`, and whitelists two CVEs which should be ignored during validation.
-It also sets `onlyFixesNotAvailable: true`, meaning that images that contain CVEs with fixes available should be denied.
 
-### 2. Deploying An Image With Vulnerabilities
-Now that we have kritis installed and an `ImageSecurityPolicy` to validate against, we can go ahead and start deploying some images!
+For more information about copying images, see [Google Cloud: Pushing and Pulling Images](https://cloud.google.com/container-registry/docs/pushing-and-pulling).
+
```
priyawadhwa: I've noticed that sometimes the vulnerability scanning is complete within a few seconds and sometimes it takes a few minutes, maybe we could add a note to wait for scanning to finish before proceeding?

tstromberg: Great idea! I've added a mention of needing to wait for the vulnerabilities to show up in gcloud after copying the image in. PTAL.

PR #207, opened 2018-08-09
Remove unnecessary use of ../ as a working directory for exec.Command invocations.

tstromberg:
Addresses lint error:

string `../` has 3 occurrences, make it a constant (goconst)


tstromberg:
Correct: I removed the strings rather than turn it into a constant, because changing the working directory is not required by these commands.

PR #211, opened 2018-08-09
K8 api  return PEM encoded value of secret.

tejal29:
kubectl get secret shows base64 encoded value of secret.
However, the k8 api fetches the secret as is without encoding it.
We no longer need to decode the secret to get to the PEM secret value.
Hence change all functions decoding the secret to use it just as is.

### DIFF
--- a/pkg/kritis/testutil/util.go
+++ b/pkg/kritis/testutil/util.go
@@ -55,25 +55,19 @@ func checkErr(shouldErr bool, err error) error {
 	return nil
 }
 
-func CreateBase64KeyPair(t *testing.T, name string) (string, string) {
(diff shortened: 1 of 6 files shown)


pkg/kritis/testutil/util.go
```diff
@@ -83,14 +77,22 @@ func getKey(key *openpgp.Entity, keyType string, t *testing.T) []byte {
 		CheckError(t, false, key.Serialize(wr))
 	}
 	wr.Close()
-	return gotWriter.Bytes()
+	return string(gotWriter.Bytes())
```
tstromberg: I believe that you should be able to skip this conversion by using: return gotWriter.String()

Ref: https://golang.org/pkg/bytes/#Buffer.String

PR #219, opened 2018-08-10
de-lint: Check return value of yaml.Unmarshal, remove type conversion, run goimports.

tstromberg:


tejal29:
restaring this build. since it failed due to some validation hooks that were around from prev builds.
Deleted all stale validation hook.

We could probably add:
```
kubectl get ValidatingWebhookConfiguration
NAME                                                                  AGE
kritis-validation-hook-deployments-ffe2ae79714bcf4a85ea1c6a7fef39f5   16h
kritis-validation-hook-ffe2ae79714bcf4a85ea1c6a7fef39f5               16h
tejaldesai@cloudshell:~ (kritis-int-test)$ kubectl delete ValidatingWebhookConfiguration --all
```

pkg/kritis/kubectl/plugins/resolve/resolve_test.go
```diff
@@ -255,9 +255,12 @@ spec:
 	testutil.CheckErrorAndDeepEqual(t, false, err, expected, actual)
 }
 
-func formatTestYaml1() yaml.MapSlice {
+func formatTestYaml1(t *testing.T) yaml.MapSlice {
 	m := yaml.MapSlice{}
```
dlorenc: Should we throw in a t.Helper() call now that this is getting a t passed in?

tejal29: i think we shd do that to all the testutil finctions. 
When debugging tests, i  always had to guess the line number where the test failed.
We should definitely call this function everwhere else.

PR #220, opened 2018-08-10
Add linter.sh to "make test" so that presubmit blocks on lint issues

tstromberg:
Depends on #219 being submitted.


tstromberg:
Yay! All dependencies have been submitted and the tests now pass. This is now ready for review.

PR #226, opened 2018-08-10
Add int tests for annotation and delete them

tejal29:
Check if annotation exits.
Delete those annotations, so for next Int Test run, we do not admit those images because they are already attested.

### DIFF
--- a/integration/crd.go
+++ b/integration/crd.go
@@ -37,7 +37,7 @@ kind: AttestationAuthority
 metadata:
   name: test-attestor
 spec:
-  noteReference: v1alpha1/projects/image-signing
+  noteReference: v1alpha1/projects/krits-int-test
   privateKeySecretName: %s
   publicKeyData: %s`
 )
--- a/integration/crd.go
+++ b/integration/crd.go
@@ -16,14 +16,34 @@ limitations under the License.
 package integration
 
 import (
+	"bytes"
+	"encoding/base64"
+	"fmt"
+	"io/ioutil"
+	"os"
 	"os/exec"
 	"testing"
 	"time"
 
 	integration_util "github.com/grafeas/kritis/pkg/kritis/integration_util"
+	"github.com/grafeas/kritis/pkg/kritis/testutil"
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
+
+const (
+	testAttesationAuthority = `apiVersion: kritis.grafeas.io/v1beta1
+kind: AttestationAuthority
+metadata:
+  name: test-attestor
+spec:
+  noteReference: v1alpha1/projects/image-signing
+  privateKeySecretName: %s
+  publicKeyData: %s`
+)
+
+// Secret name for test-attestor
+var aaSecret = "test-attestor"
 
 // CRDs is a map of CRD type to names of the expected CRDs to create.
 var crdNames = map[string]string{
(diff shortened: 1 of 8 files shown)


integration/run_test.go
```diff
@@ -442,7 +460,7 @@ func TestKritisPods(t *testing.T) {
 				t.Fatalf("deployment should have failed but succeeded\n%s",
 					getKritisLogs(t))
 			}
-
+			checkAndDeleteAttestations(t, testCase.attestedImages)
```
tstromberg: I don't understand what this is testing at all. Can you refactor this so that the comparisons are done within the test function and not in a helper?

"Readers should need to read only the test body to understand the test". 

See also: 

https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures
https://g3doc.corp.google.com/go/g3doc/go-readability-review-comments.md?cl=head#TOC-Assert

pkg/kritis/metadata/metadata.go
```diff
@@ -48,4 +48,5 @@ type Vulnerability struct {
 type PGPAttestation struct {
 	Signature string
 	KeyID     string
+	OccName   string
```
tstromberg: I don't know what an OccName is. Do you mind adding a docstring so that this shows up in our autogenerated docs properly?

integration/run_test.go
tejal29: sure. will do.

tejal29:
@tstromberg this is now ready for review!

integration/crd.go
```diff
@@ -16,38 +16,90 @@ limitations under the License.
...
+	cmd := exec.Command("kubectl", "create", "secret", "generic", aaSecret,
+		fmt.Sprintf("--from-file=public=%s", pubFile),
+		fmt.Sprintf("--from-file=private=%s", privFile),
+		"-n", ns)
+	if _, err := integration_util.RunCmdOut(cmd); err != nil {
+		t.Fatalf("unexpected error %s", err)
+	}
+
+	// Create the Attestation authority
+	createAA(t, ns, pubKeyEnc)
```
tstromberg: I'm not going to ask you to fix this all now, but we should never be at the point where we are nesting test helpers. I'm less strict about integration tests versus unit tests, but I believe that we are now heading down a path that obscures the connection between the test failure and the conditions which led to it. Some good rules of thumb:

- Test helper functions should never call other test helper functions. Doing so makes it difficult to follow the code path. Test readability is more critical than code readability.

- Test helpers should not interact with the code under test. This helper does not, but getAttestations() seems to fall into this category.

- Never pass testing.T into functions when the standard Go pattern of returning an error is sufficient. 

More context: http://go/gocomments#TOC-Assert

integration/run_test.go
```diff
@@ -224,8 +225,9 @@ func TestKritisISPLogic(t *testing.T) {
 					name: "nginx-no-digest",
 				},
 			},
-			shouldSucceed: false,
-			dir:           "../",
+			shouldSucceed:  false,
+			attestedImages: []string{},
+			dir:            "../",
```
tstromberg: Heads up: The dir attribute will be removed by https://github.com/grafeas/kritis/pull/237

tejal29: thanks. will get my PR in first :)

integration/crd.go
tejal29: ahh. thanks! we do that many places everywhere to reduce duplication and work around checking error and failing.

Issue #233, opened 2018-08-10
Kritis Bug bash Issues.


tejal29:
Thanks @sharifelgamal for testing the tutorial. 
Here is the feedback
 - [ ] Mention in Bold installing Kritis as a pre requirement for tutorial.
- [ ] Installing kritis mentions helm as a requirement and then users do the `helm init` as part of installation steps on the helm installation guide. We have the right helm installation steps required  https://github.com/grafeas/kritis/blob/master/install.md#step-6-install-and-configure-helm. Lets remove the helm from requirement.
- [ ] Opening all google Enable API links in new window. Or we can also enable API from command line. https://cloud.google.com/endpoints/docs/openapi/enable-api Lets use that.
- [ ] Right plugin install instructions from  #231
- [ ] helm install kritis-charts should point to  GCS bucket.
- [ ] the pod used in example, crashes. Have a pod which does not crash or atleast ask users to ignore it
- [ ] Link Delete kritis from Tutorial at the end
- [ ] The tutorial shd have instructions to delete all the pods that are created.

priyawadhwa:
I figured the last two could be combined, added in #236

PR #239, opened 2018-08-13
update tutorial with image digest which doesn't crash

### LINKED ISSUE #235: Create a image which does not crash for tutorial.
The image used to create a pod in  tutorial, gcr.io/kritis-tutorial/java-with-vuln:latest crashes.
We should create a pod which does not crash.

priyawadhwa:
Should fix #235 and part of #233 

I added an entrypoint to the image so that it would run forever


tejal29:
Shd we also use  the same image in the integration test examples?
That way our tests will also test the images listed in the tutorial

integration/testdata/java/java-with-vuln.yaml
```diff
@@ -1,10 +0,0 @@
-apiVersion: v1
-kind: Pod
-metadata:
-  name: java-with-vuln
-spec:
-  containers:
-  - name: java-with-vuln
-    image: gcr.io/kritis-int-test/java-with-vuln@sha256:b3f3eccfd27c9864312af3796067e7db28007a1566e1e042c5862eed3ff1b2c8
-    ports:
-    - containerPort: 80
```
priyawadhwa: this file wasn't deleted just renamed (not sure why github isn't showing that)

tejal29: yeah strange.  it should show an rename and a change.

PR #244, opened 2018-08-13
Create v0.1.0 release

tejal29:
Followed the steps on https://github.com/grafeas/kritis/blob/master/RELEASING.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 This is the initial release of Kritis
+
 For Release v0.1.0:
   * Kritis is only available on GKE.
   * Kritis allows users to secure their clusters by specifying the vulnerability threshold for containers above which containers cannot be deployed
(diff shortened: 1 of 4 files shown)


kritis-charts/values.yaml
```diff
@@ -7,7 +7,7 @@ replicaCount: 1
 image:
   image: kritis-server
   # TODO(tstromberg): Have a versioned tag after release.
```
priyawadhwa: We could probably get rid of this TODO, @tstromberg can confirm!

CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+This is the initial release of Kritis
+For Release v0.1.0:
```
priyawadhwa: nit: newline here, it's appearing on the same line as the previous sentence

kritis-charts/values.yaml
tejal29: will remove.

Issue #250, opened 2018-08-15
check for "Controlled By" section before admitting pods


priyawadhwa:
Right now if we validate and admit a deployment, we validate the pods it creates again. 

It looks like if pods are created by other objects (Deployment, ReplicaSet etc) then there's a "Controlled By" section in the pod description which we could check for before validation. If the "Controlled By" section exists, we can let the pod through (assuming whatever created it was already validating and approved)

```shell
$ kubectl describe pod test-5d54fc85c-bn72b 
Name:           test-5d54fc85c-bn72b
Namespace:      default
Node:           gke-kritis-default-pool-f1c1e77c-n3sw/10.128.0.2
Start Time:     Wed, 15 Aug 2018 10:19:01 -0700
Labels:         pod-template-hash=181097417
                run=test
Annotations:    kubernetes.io/limit-ranger=LimitRanger plugin set: cpu request for container test
Status:         Running
IP:             10.64.2.52
Controlled By:  ReplicaSet/test-5d54fc85c
```

For this to work, we'll need to add:
1. ReplicaSet type to kritis, since Deployments create ReplicaSets which in turn create Pods
2. The actual logic for checking the "Controlled By" section

tejal29:
Nice!

PR #251, opened 2018-08-15
Add replica set type

priyawadhwa:
First part of fixing #250 is to add support for validating ReplicaSets to kritis

### DIFF
--- a/pkg/kritis/admission/admission.go
+++ b/pkg/kritis/admission/admission.go
@@ -69,6 +69,7 @@ var (
 var handlers = map[string]func(*v1beta1.AdmissionReview, *v1beta1.AdmissionReview) error{
 	"Deployment": handleDeployment,
 	"Pod":        handlePod,
+	"ReplicaSet": handleReplicaSet,
 }
 
 func handleDeployment(ar *v1beta1.AdmissionReview, admitResponse *v1beta1.AdmissionReview) error {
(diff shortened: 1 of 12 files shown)


pkg/kritis/admission/admission.go
```diff
@@ -91,6 +92,16 @@ func handlePod(ar *v1beta1.AdmissionReview, admitResponse *v1beta1.AdmissionRevi
 	return nil
 }
 
+func handleReplicaSet(ar *v1beta1.AdmissionReview, admitResponse *v1beta1.AdmissionReview) error {
+	glog.Info("handling replica set...")
```
tstromberg: As an operator, I don't think I would find this message useful unless it mentioned which replica set it is handling. I suggest either adding more context, or using glog.V(2).Info.

priyawadhwa: for sure, added more info to this error and the ones for deployment/pods

pkg/kritis/admission/admission.go
```diff
@@ -225,6 +236,20 @@ func reviewPod(pod *v1.Pod, ar *v1beta1.AdmissionReview) {
 	reviewImages(pods.Images(*pod), pod.Namespace, pod, ar)
 }
 
+func reviewReplicaSet(replicaSet *appsv1.ReplicaSet, ar *v1beta1.AdmissionReview) {
+	// First, check for a breakglass annotation on the replica set
+	if checkBreakglass(&replicaSet.ObjectMeta) {
+		glog.Infof("found breakglass annotation for %s, returning successful status", replicaSet.Name)
+		return
+	}
+	for _, c := range replicaSet.Spec.Template.Spec.InitContainers {
```
tejal29: How about we collect all images together. That ways, the reviewImages will prune duplicate images.
Calling them one by one would not do that.

tejal29: Can we add unit tests for this?

priyawadhwa: Done!

pkg/kritis/admission/images.go
```diff
@@ -0,0 +1,58 @@
...
+	for _, ic := range pod.Spec.InitContainers {
+		images = append(images, ic.Image)
+	}
+	for _, c := range pod.Spec.Containers {
+		images = append(images, c.Image)
+	}
+	return images
+}
+
+// DeploymentImages returns a list of images in a pod
```
tstromberg: This docstrings for these 3 functions should be different from one another.

priyawadhwa: done

PR #258, opened 2018-08-21
Check owners of a Kubernetes type before validation

### LINKED ISSUE #250: check for "Controlled By" section before admitting pods

priyawadhwa:
This PR adds support for checking if an object's owner has been
validated before starting validation. If an object is owned by a type
supported by Kritis (currently Pods, Deployments and ReplicaSets) then
we know that it has already been validated and we don't have to validate
again.

This should fix #250


dlorenc:
How safe is this? Can someone set an owner reference after the fact?

priyawadhwa:
So it looks like you can set it after the fact, from [here](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents)
>You can also specify relationships between owners and dependents by manually setting the ownerReference field.

Theoretically you can set it to anything, even something that doesn't exist. If we make sure that:
1. An object's owner actually exists
2. The object contains the same images as the owner (or a subset of the owner's images)

would that be enough to allow an object?

tejal29:
Is there a way to get the name of the owner object? If that is the case, your suggestion 1 to check  if owner exists should work. 
For pt 2, we need to think about images with tags.

pkg/kritis/admission/admission.go
```diff
@@ -272,6 +285,22 @@ func unmarshalDeployment(r *http.Request) (*appsv1.Deployment, v1beta1.Admission
...
 }
 
+func checkOwners(meta *metav1.ObjectMeta) bool {
+	owners := meta.GetOwnerReferences()
+	if owners == nil {
+		return false
+	}
+	for _, o := range owners {
+		for _, s := range constants.SupportedTypes {
+			if o.Kind == s {
```
tejal29: add a check if onwer is running like we mentioned.

pkg/kritis/admission/admission.go
```diff
@@ -178,11 +178,17 @@ func ReviewHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func reviewDeployment(deployment *appsv1.Deployment, ar *v1beta1.AdmissionReview) {
+	images := DeploymentImages(*deployment)
+	// check if the Deployments's owner has already been validated
+	if checkOwners(images, &deployment.ObjectMeta) {
+		return
```
tejal29: Can we log here and elsewhere so its easier to know why the object was admitted.

PR #268, opened 2018-08-30
Add caching for CA client

tejal29:


pkg/kritis/metadata/containeranalysis/cache.go
```diff
@@ -0,0 +1,94 @@
...
+	kritisv1beta1 "github.com/grafeas/kritis/pkg/kritis/apis/kritis/v1beta1"
+	"github.com/grafeas/kritis/pkg/kritis/metadata"
+	"github.com/grafeas/kritis/pkg/kritis/secrets"
+	containeranalysispb "google.golang.org/genproto/googleapis/devtools/containeranalysis/v1alpha1"
+)
+
+// Cache struct defines Cache for container analysis client.
+type Cache struct {
+	client metadata.Fetcher
+	vCache map[string][]metadata.Vulnerability
```
tstromberg: Do you mind naming this something different? Cache.*Cache stutters, and includes words that are already clear given the surrounding context. How about just cache.Vuln or cache.Vulnerability?

Same with the other fields within this struct.

pkg/kritis/metadata/containeranalysis/cache.go
```diff
@@ -0,0 +1,94 @@
...
+	return &Cache{
+		client: c,
+		vCache: map[string][]metadata.Vulnerability{},
+		aCache: map[string][]metadata.PGPAttestation{},
+		nCache: map[*kritisv1beta1.AttestationAuthority]*containeranalysispb.Note{},
+	}, nil
+}
+
+// GetVulnerabilities gets Package Vulnerabilities Occurrences for a specified image.
+func (c Cache) GetVulnerabilities(image string) ([]metadata.Vulnerability, error) {
```
tstromberg: optional: The "Get" in function names is implicit, and makes the name longer, so Go it'd probably be more idiomatic to omit it. That said, it's more important to be consistent with the other parts of the code base, so feel free to ignore this suggestion.

pkg/kritis/metadata/containeranalysis/cache.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	n, err := c.client.GetAttestationNote(aa)
+	if err != nil {
+		c.nCache[aa] = n
+	}
+	return n, err
+}
+
+// CreateAttestationOccurence creates an Attestation occurrence for a given image and secret.
+func (c Cache) CreateAttestationOccurence(n *containeranalysispb.Note,
```
tstromberg: Can you remove the newlines from the function declaration? It seems awkward, and can be solved in other ways. See: https://github.com/golang/go/wiki/CodeReviewComments#line-length

pkg/kritis/metadata/containeranalysis/cache.go
```diff
@@ -0,0 +1,94 @@
...
+limitations under the License.
+*/
+
+package containeranalysis
+
+import (
+	kritisv1beta1 "github.com/grafeas/kritis/pkg/kritis/apis/kritis/v1beta1"
+	"github.com/grafeas/kritis/pkg/kritis/metadata"
+	"github.com/grafeas/kritis/pkg/kritis/secrets"
+	containeranalysispb "google.golang.org/genproto/googleapis/devtools/containeranalysis/v1alpha1"
```
tstromberg: optional: Naming this import "cpb" would result in shorter function declarations, and would be more idiomatic.

pkg/kritis/metadata/containeranalysis/cache.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		c.nCache[aa] = n
+	}
+	return n, err
+}
+
+// CreateAttestationOccurence creates an Attestation occurrence for a given image and secret.
+func (c Cache) CreateAttestationOccurence(n *containeranalysispb.Note,
+	image string,
+	pKey *secrets.PGPSigningSecret) (*containeranalysispb.Occurrence, error) {
```
tstromberg: That this is a key is obvious given the context. A single character variable name should suffice here.

pkg/kritis/metadata/containeranalysis/cache_test.go
```diff
@@ -0,0 +1,135 @@
...
+		vCache: nil,
+		aCache: map[string][]metadata.PGPAttestation{"image-hit": pgpAtt},
+		nCache: nil,
+	}
+	tcs := []struct {
+		name     string
+		image    string
+		expected []metadata.PGPAttestation
+	}{
+		{"hit", "image-hit", pgpAtt},
```
tstromberg: Inline pgpAtt: Readers should need to read only the test body to understand the test. Removing this minor bit of duplication is not worth the additional cognitive overhead for future readers.

pkg/kritis/metadata/containeranalysis/cache_test.go
```diff
@@ -0,0 +1,135 @@
...
+		nCache: nil,
+	}
+	tcs := []struct {
+		name     string
+		image    string
+		expected []metadata.Vulnerability
+	}{
+		{"hit", "image-hit", vuln},
+		{"miss", "image-miss", nil},
+	}
```
tstromberg: Can you add a test where a cache miss still results in valid data? 

It's difficult to tell otherwise whether or not this package accesses authoritative data in the case of a cache miss, or simply returns nil.

pkg/kritis/metadata/containeranalysis/cache_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+var note = &containeranalysispb.Note{
+	Name: "test-note",
+}
+var mock = &testutil.MockMetadataClient{}
+
+func TestVCache(t *testing.T) {
+	c := Cache{
+		client: mock,
+		vCache: map[string][]metadata.Vulnerability{"image-hit": vuln},
```
tstromberg: Inline vuln: Readers should need to read only the test body to understand the test. Removing this minor bit of duplication is not worth the additional cognitive overhead for future readers.

If you feel like it's too long, you can still test with a simpler object without newlines, like []metadata.Vulnerability{{CVE: "test"}}

pkg/kritis/metadata/containeranalysis/cache.go
tejal29: yes. makes sense.

pkg/kritis/metadata/containeranalysis/cache.go
tejal29: done

pkg/kritis/metadata/containeranalysis/cache.go
tejal29: done here and elsewhere.

pkg/kritis/metadata/containeranalysis/cache.go
tejal29: done

pkg/kritis/metadata/containeranalysis/cache_test.go
tejal29: done

pkg/kritis/metadata/containeranalysis/cache_test.go
tejal29: yes will do

pkg/kritis/metadata/containeranalysis/cache_test.go
tejal29: done

tejal29:
@tstromberg fixed.

PR #269, opened 2018-08-30
DO NOT REVIEW: Refactor integration test to fully respect --gcp-project

tstromberg:


integration/cleanup.go
```diff
@@ -19,29 +19,37 @@ package integration
...
+		Type string
+		Name string
+	}{
+		{"validatingwebhookconfiguration", fmt.Sprintf("kritis-validation-hook-%s", ns.Name)},
+		{"validatingwebhookconfiguration", fmt.Sprintf("kritis-validation-hook-deployments-%s", ns.Name)},
+		{"csr", fmt.Sprintf("tls-webhook-secret-cert-%s", ns.Name)},
+		{"secret", aaSecret},
+	}
+	for _, r := range resources {
+		if err := deleteObject(r.Type, r.Name, ns); err != nil {
```
tejal29: The first 3 resources  in the list are installed in default namespace and only `secret` is installed in test namespace

tstromberg: Great catch. Fixed this, and the fact that the code was not checking the error code from cleanupInstall.

PR #271, opened 2018-08-31
Makefile: allow integration testers to pass in project & cluster values.

tstromberg:
For example:

make -e GCP_PROJECT=my-project -e TEST_CLUSTER=test integration-test

This also adds a "setup-integration-local" rule, which does everything
necessary to setup a GCP cluster to run integration tests.

One requirement that came out during testing is that GAC_CREDENTIALS_PATH is now stored with a path that allows credentials to be stored for multiple projects.

NOTE: Requires PR #269 to be submitted, and helps to address issue #237.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,11 @@ COMMIT ?= $(shell git rev-parse HEAD)
 VERSION ?= v0.1.0
 IMAGE_TAG ?= $(COMMIT)
 
-# TODO(aaron-prindle) add this env var for int-test configuration
-# GCP_TEST_PROJECT ?= kritis-int-test
+# Used for integration testing. example:
+# "make -e GCP_PROJECT=kritis-int integration-local"
+GCP_PROJECT ?= kritis-int
+GCP_ZONE ?= us-central1-a
+TEST_CLUSTER ?= test-cluster-2
 
 %.exe: %
 	mv $< $@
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -20,26 +20,30 @@ COMMIT ?= $(shell git rev-parse HEAD)
 VERSION ?= v0.1.0
 IMAGE_TAG ?= $(COMMIT)
 
-# TODO(aaron-prindle) add this env var for int-test configuration
-# GCP_TEST_PROJECT ?= kritis-int-test
+# Used for integration testing. example:
+# "make -e GCP_PROJECT=kritis-int integration-local"
+GCP_PROJECT ?= kritis-int-test
```
tejal29: maybe this should be TEST_GCP_PROJECT or just TEST_PROJECT?

Makefile
```diff
@@ -20,26 +20,30 @@ COMMIT ?= $(shell git rev-parse HEAD)
...
+GCP_ZONE ?= us-central1-a
```
tejal29: same here

Makefile
```diff
@@ -172,3 +180,55 @@ gcb-signer-image: out/gcb-signer-image
 gcb-signer-push-image: gcb-signer-image
 	docker push $(REGISTRY)/kritis-gcb-signer:$(IMAGE_TAG)
 
+# Steps to create a new test cluster on an GCP project that has already had a Kritis setup.
+.PHONY: create-integration-local
+create-integration-local:
+	gcloud --project=$(GCP_PROJECT) container clusters describe $(TEST_CLUSTER) >/dev/null \
```
tejal29: does this deserve to be in own script file?

Makefile
```diff
@@ -172,3 +180,55 @@ gcb-signer-image: out/gcb-signer-image
...
+	gcloud -q container images add-tag \
+		gcr.io/kritis-tutorial/nginx-no-digest-breakglass:latest \
+		gcr.io/$(GCP_PROJECT)/nginx-no-digest-breakglass:latest
+	gcloud -q container images add-tag \
+		gcr.io/kritis-tutorial/nginx-no-digest:latest \
+		gcr.io/$(GCP_PROJECT)/nginx-no-digest:latest
+
+# Fully setup local integration testing.
+.PHONY: setup-integration-local
+setup-integration-local: build-push-test-image create-integration-local
```
tejal29: I think build-push-test-image should be moved over to integration local.
I want to iterate over kritis code and run int test,  running setup-integration-local multiple times will create an error for  already set up cluster.
Ideally, i would want to my dev flow to be: 
```
make setup-integration-local 

make integration-local
make integration-local 
make integration-local
...
```
Right now,  i will have to 
```
make setup-integration-local
make build-push-test-image integration-local

```

Makefile
```diff
@@ -172,3 +180,55 @@ gcb-signer-image: out/gcb-signer-image
...
 	docker push $(REGISTRY)/kritis-gcb-signer:$(IMAGE_TAG)
 
+# Steps to create a new test cluster on an GCP project that has already had a Kritis setup.
+.PHONY: create-integration-local
+create-integration-local:
+	gcloud --project=$(GCP_PROJECT) container clusters describe $(TEST_CLUSTER) >/dev/null \
+		|| gcloud --project=$(GCP_PROJECT) container clusters create $(TEST_CLUSTER) \
+		--num-nodes=2 --zone=$(GCP_ZONE)
+	gcloud --project=$(GCP_PROJECT) container clusters get-credentials $(TEST_CLUSTER)
+	mkdir -p $(dir $(GAC_CREDENTIALS_PATH))
```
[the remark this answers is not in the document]
tstromberg: It will not fail in this case:

       -p, --parents
              no error if existing, make parent directories as needed

Makefile
tstromberg: I'm on the fence here - adding another level of indirection doesn't have a clear benefit for me, but I can do so if you disagree.

Makefile
tstromberg: Good idea. Done!

PR #275, opened 2018-09-01
Include output of get-credentials when encountering an error.

tstromberg:


integration/run_test.go
```diff
@@ -240,8 +240,9 @@ func setUp(t *testing.T) (kubernetes.Interface, *v1.Namespace, func(t *testing.T
 	}
 
 	cmd := exec.Command("gcloud", "container", "clusters", "get-credentials", *gkeClusterName, "--zone", *gkeZone, "--project", *gcpProject)
-	if err := integration_util.RunCmd(cmd); err != nil {
```
tejal29: you can still inline it. 
```
if out, err := integration_util.RunCmd(cmd); err != nil{
t.Fatalf("get-credentials: %v - %s\n\nPlease ensure that \"make setup-integration-local\" has been run first", out, err)
}
```

PR #276, opened 2018-09-06
Refactor Violation as a interface

tejal29:
This is first step of Refactoring Kritis Policy Engine.

Refactoring Violation as policy.Violation interface. 
This will allow other policies like Build Policy to specify their own Violation structs.
The violation.Strategy can now handle ISP as well as Build Policy Violations (in future).

### DIFF
--- a/pkg/kritis/crd/securitypolicy/securitypolicy_test.go
+++ b/pkg/kritis/crd/securitypolicy/securitypolicy_test.go
@@ -76,12 +76,12 @@ func Test_UnqualifiedImage(t *testing.T) {
 		},
 	}
 	violations, err := ValidateImageSecurityPolicy(isp, "", &testutil.MockMetadataClient{})
-	testutil.CheckError(t, false, err)
-	if len(violations) != 1 {
-		t.Errorf("expected only 1 violation. Got %d", len(violations))
-	}
-	testutil.CheckErrorAndDeepEqual(t, false, err, violations[0].Type(), policy.UnqualifiedImageViolation)
-	testutil.CheckErrorAndDeepEqual(t, false, err, violations[0].Reason(), UnqualifiedImageReason(""))
+	expected := []policy.Violation{}
+	expected = append(expected, Violation{
+		vType:  policy.UnqualifiedImageViolation,
+		reason: UnqualifiedImageReason(""),
+	})
+	testutil.CheckErrorAndDeepEqual(t, false, err, expected, violations)
 }
 
 func Test_SeverityThresholds(t *testing.T) {
--- a/pkg/kritis/crd/securitypolicy/securitypolicy_test.go
+++ b/pkg/kritis/crd/securitypolicy/securitypolicy_test.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/grafeas/kritis/pkg/kritis/apis/kritis/v1beta1"
 	"github.com/grafeas/kritis/pkg/kritis/metadata"
+	"github.com/grafeas/kritis/pkg/kritis/policy"
 	"github.com/grafeas/kritis/pkg/kritis/testutil"
 )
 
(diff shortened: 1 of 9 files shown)


pkg/kritis/crd/securitypolicy/securitypolicy_test.go
```diff
@@ -75,14 +76,12 @@ func Test_UnqualifiedImage(t *testing.T) {
...
-			Vulnerability: metadata.Vulnerability{},
-			Violation:     UnqualifiedImageViolation,
-			Reason:        UnqualifiedImageReason(""),
-		},
+	testutil.CheckError(t, false, err)
+	if len(violations) != 1 {
+		t.Errorf("expected only 1 violation. Got %d", len(violations))
 	}
-	testutil.CheckErrorAndDeepEqual(t, false, err, violations, expected)
+	testutil.CheckErrorAndDeepEqual(t, false, err, violations[0].Type(), policy.UnqualifiedImageViolation)
```
tstromberg: Would it be possible to compare the result against against expectations and showing a diff for all the fields that don't match? It's generally recommended in Go as a way to make tests easier to reason about and maintain.

Context: http://go/go-test-comments#compare-full-structures and http://go/go-test-comments#print-diffs

pkg/kritis/crd/securitypolicy/violation.go
```diff
@@ -22,48 +22,54 @@ import (
...
-	SeverityViolation
+	"github.com/grafeas/kritis/pkg/kritis/policy"
 )
 
 // Violation represents a vulnerability that violates an ISP
 type Violation struct {
 	Vulnerability metadata.Vulnerability
-	Violation     int
-	Reason        Reason
+	VType         policy.ViolationType
```
tstromberg: Do you mind finding less of a stuttering name than violation.VType, such as violation.Type?

pkg/kritis/crd/securitypolicy/violation.go
```diff
@@ -22,48 +22,54 @@ import (
...
+	Msg           policy.Reason
```
tstromberg: What's the background behind the terminology change here? If possible, I'd rather that the Policy & Violation structs use the same terminology.

pkg/kritis/crd/securitypolicy/violation.go
```diff
@@ -22,48 +22,54 @@ import (
...
+}
+
+// Reason returns the reason
+func (v Violation) Reason() policy.Reason {
```
tstromberg: Public getters for public struct fields seems unusual. Structs should either have public fields, or public getters, not both. 

In this particular case, I don't understand why we introduce the additional overhead of implementing getters for a perfectly usable public field, unless there are bigger plans here.

pkg/kritis/review/review_test.go
```diff
@@ -131,16 +132,17 @@ func TestReview(t *testing.T) {
 				PublicKeyData:        sec.PublicKey,
 			}}}, nil
 	}
-	testValidate := func(isp v1beta1.ImageSecurityPolicy, image string, client metadata.Fetcher) ([]securitypolicy.Violation, error) {
+	testValidate := func(isp v1beta1.ImageSecurityPolicy, image string, client metadata.Fetcher) ([]policy.Violation, error) {
```
tstromberg: By the name here, I assumed this was implementing a test, and I didn't quite understand it. Should this be instead named mock or fake?

pkg/kritis/crd/securitypolicy/violation.go
tejal29: Type is a keyword and i could not use it. hence i named is vType.

tejal29: i can double check this. because last i tried, i was using `type` vs 'Type`

pkg/kritis/crd/securitypolicy/violation.go
tejal29: Violation.Violation sounded a little bit weird.
I had to change Reason -> Msg because you can't have a same name for method and variable.

pkg/kritis/review/review_test.go
tejal29: sure will do

pkg/kritis/crd/securitypolicy/violation.go
tejal29: Since this struct implements an interface i did this. 
I was debating if i should make  the fields private and then add a New function (since its used across package boundaries). i can do that.

pkg/kritis/crd/securitypolicy/violation.go
tejal29: yes! `Type` works but `type` does not. Since i changed these to be a private variables i renamed it as vType.

PR #283, opened 2018-09-18
Set the PGP key fingerprint as the key id for attestations.

wteiken:
Currently the secret name is set.   The attestation spec requires this to be the fingerprint of the PGP key used for signing.


wteiken:
This should break the review code, but it seems the integration tests don't detect this.

wteiken:
@tejal29 friendly ping

tejal29:
Sorry about the delay @wteiken

PR #288, opened 2018-10-20
remove "Your account must be whitelisted to enable the Container Anal…

leewalter:
…ysis API."

I tested it on 2018-10-19 on GCP and this is available as BETA and no need for account whitelisting and wait as said earlier , "It may take 1-5 business days to approve the request."

so, suggest to remove this few lines which is likely obsolete now.

PR #289, opened 2018-10-20
Leewalter tip gpg hung1

leewalter:
try to add a tip when gpg hung at quick-create-key because of insufficient entropy in Ubuntu on GCP.
It also has another suggestion in PR 
https://github.com/grafeas/kritis/pull/288

### DIFF
--- a/install.md
+++ b/install.md
@@ -33,9 +33,7 @@ gcloud projects list
 
 ## Step #2: Enable the requisite API's for your Google Cloud Project
 
-NOTE: Your account must be whitelisted to enable the Container Analysis API. To do so, join the  [Container Analysis Users Group](https://groups.google.com/forum/#!forum/containeranalysis-users). It may take 1-5 business days to approve the request.
-
-Once approved, enable the necessary API's:
+Enable the necessary API's:
 
 Enable the Container Analysis API:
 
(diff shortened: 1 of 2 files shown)


tutorial.md
```diff
@@ -44,6 +44,8 @@ Note: Please create a key with Empty Passphase. We are working on adding support
 ```shell
 gpg --quick-generate-key --yes my.attestator@example.com
 
+(if you use Ubuntu on GCP and got hung in above, then please see https://delightlylinux.wordpress.com/2015/07/01/is-gpg-hanging-when-generating-a-key/ to make sure you got enough entropy.)
```
tejal29: usually, people just copy all the commands in the shell block and  paste them. 
They will blindly copy your amazing Tip and it will error out on the terminal.
Can you please move the tip after the shell block ends @L52 or
 1. end the shell block after `gpg --quick-generate-key` at L47. Add you tip and the begin the shell for the following command at L48

install.md
```diff
@@ -33,9 +33,7 @@ gcloud projects list
 
 ## Step #2: Enable the requisite API's for your Google Cloud Project
 
-NOTE: Your account must be whitelisted to enable the Container Analysis API. To do so, join the  [Container Analysis Users Group](https://groups.google.com/forum/#!forum/containeranalysis-users). It may take 1-5 business days to approve the request.
```
tejal29: Thank you! Great! you able to to use account without the whitelist?

tejal29: just read your comment in other PR. Looks like you were able too. Thanks for deleting this!

PR #290, opened 2018-10-21
Add grafeas support - v1

liron-l:
## Changes
1. Implements the `Fetcher` API using grafeas beta client (remark:
grafeas project has recently gone migration from v1alpha1 to b1beta1,
grafeas/grafeas#220)
2. Move common functions to `util.go`
## Solution caveats
The current implementation has the follow caveats:
    1. Only a single project is supported
    2. Connection to grafeas is only via unix socket (supported after
grafeas/grafeas#220)

## Next steps:
1. Enable specifying a grafeas setting config (client certs and URL)
2. More tests and validations
3. Add filtering capabilities to grafeas to support existing kritis queries
(resource url and kind). This will be done in grafeas project.

### DIFF
--- a/pkg/kritis/admission/admission.go
+++ b/pkg/kritis/admission/admission.go
@@ -22,6 +22,9 @@ import (
 	"io/ioutil"
 	"net/http"
 
+	"github.com/grafeas/kritis/pkg/kritis/metadata/containeranalysis"
+	"github.com/grafeas/kritis/pkg/kritis/metadata/grafeas"
+
 	"github.com/golang/glog"
 	"github.com/grafeas/kritis/cmd/kritis/version"
 	"github.com/grafeas/kritis/pkg/kritis/admission/constants"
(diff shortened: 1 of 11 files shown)


tejal29:
Thanks @liron-l for this change!
Right now, only the cron job uses the backend config value.
We also need to make the same change and pass the config to admission controller.  You will have to remove the hard coding here.
https://github.com/grafeas/kritis/blob/master/pkg/kritis/admission/admission.go#L309

liron-l:
Thanks @tejal29
1. I've added support for specifying configuration to the admission controller handlers. 
Please let me know if you have another pattern in mind for injecting the configuration to the handlers.
2. I think the webhooks are disabled due to GH [status](https://status.github.com/messages)

pkg/kritis/admission/admission.go
```diff
@@ -306,7 +307,10 @@ func checkBreakglass(meta *metav1.ObjectMeta) bool {
 }
 
 // TODO: update this once we have more metadata clients
```
tejal29: Remove this TODO.
Shd we use the same function for getting the client in cron.go and admission.go and remove the code duplication.

liron-l: @tejal29 good suggestion. I updated the PR.

PR #295, opened 2018-10-25
(NEEDS WORK) Implement lint changes recommended by gometalinter

tstromberg:
It turns out that golangci-lint was hiding some issues that the gometalinter work surfaced. Here's the issues that were raised.

Upgrading the golangci-lint version and configuration (#305) to match will be in a subsequent PR.


jirfag:
hi!
we've changed [license](https://github.com/golangci/golangci-lint/blob/master/LICENSE) to GPL. No need to switch :)

tejal29:
Shd we discard this? @tstromberg

tstromberg:
The lint changes are still reasonable, so I'll modify this PR to upgrade to the latest golangci-lint with a similar configuration.

PR #296, opened 2018-10-25
Clarified details about namespace and release names

aysylu:
When uninstalling Kritis.

### DIFF
--- a/install.md
+++ b/install.md
@@ -230,7 +230,9 @@ Then delete the name of the release:
 helm delete <name>
 ```
 
-This command will also kick off the `kritis-predelete` pod, which deletes the CertificateSigningRequest, TLS Secret, and Webhooks created during installation. You may view the status using:
+`name` in this case is `loopy-numbat`.
+
+You don'tThis command will also kick off the `kritis-predelete` pod, which deletes the CertificateSigningRequest, TLS Secret, and Webhooks created during installation. You may view the status using:
 
 ```shell
 kubectl get pods kritis-predelete
@@ -249,6 +251,8 @@ To delete the remaining resources run:
 ```
 kubectl delete pods,serviceaccount,clusterrolebinding --selector kritis.grafeas.io/install --namespace <your namespace>
 ```
+
+If you used the default namespace during the installation, you don't need to provide the `--namespace <your namespace>`.
 
 NOTE: This will not delete the container analysis secret created above.
 


install.md
```diff
@@ -230,7 +230,9 @@ Then delete the name of the release:
 helm delete <name>
 ```
 
-This command will also kick off the `kritis-predelete` pod, which deletes the CertificateSigningRequest, TLS Secret, and Webhooks created during installation. You may view the status using:
+`name` in this case is `loopy-numbat`.
+
+You don'tThis command will also kick off the `kritis-predelete` pod, which deletes the CertificateSigningRequest, TLS Secret, and Webhooks created during installation. You may view the status using:
```
tstromberg: Remove "You don't".

install.md
```diff
@@ -250,6 +252,8 @@ To delete the remaining resources run:
 kubectl delete pods,serviceaccount,clusterrolebinding --selector kritis.grafeas.io/install --namespace <your namespace>
 ```
 
+If you used the default namespace during the installation, you don't need to provide the `--namespace <your namespace>`.
```
tstromberg: minor nit - feel free to disagree/ignore: I feel like this could be improved so that:

- We don't ask users to rewrite the command line they've pasted based on something they may not be sure about.
- We clarify before they are likely to paste the command in.

Two ideas:

"To delete the remaining resources, run the following, where `<namespace>` is `default`, or the name you specifically selected at installation time:
..."

Or perhaps mentioning afterwards in such a way that simply has them plug the appropriate value into --namespace rather than omitting it:

If you did not specifically select a namespace during installation, the default value is `default`.

Thoughts?

install.md
```diff
@@ -250,6 +252,8 @@ To delete the remaining resources run:
 kubectl delete pods,serviceaccount,clusterrolebinding --selector kritis.grafeas.io/install --namespace <your namespace>
```
tstromberg: Can you break this command apart so that it fits in a browser window?

```
kubectl delete pods,serviceaccount,clusterrolebinding \
  --selector kritis.grafeas.io/install \
  --namespace <your namespace>
```

aysylu: done

install.md
aysylu: That's a great point. I went with your second suggestion, as I think it sheds some light into the internals that weren't mentioned before, while keeping it as a side note.

install.md
aysylu: Done

Issue #224, opened 2018-08-10
[Idea] Create a new CRD for Kritis Configuration Policy.


tejal29:
This CRD will define a few configs for Kritis like

1. Which metadata api to use
2. What shd be the default fallback strategy in case Metadata API is not available.
3. Default Cron job run interval.
And anything else which is hardcoded.
This CRD will be cluster scoped for now meaning it will apply to the whole cluster.

The controller and cron job will first fetch this CRD use that instead of defaults.

Pros
1. All main falgs can be customized
2. cluster Admins can apply new config changes without needing to restart kritis.
3. Easy to switch back and forth between  Metadata api
4. Easy to change fallback policy.

tejal29:
Tutorial for adding 
1. CRDs is [here](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition)  
2. code generation for Custom Resources is [here](https://blog.openshift.com/kubernetes-deep-dive-code-generation-customresources/)

Now lets add a new custom resource definition
1. Define a new Custom Resource Definition for for ` KritisConfig` like [this](https://github.com/grafeas/kritis/blob/493cd0d4f73bfc6aa8804951c1e6cc8da380c336/artifacts/build-policy-crd.yaml) lets say you name is as `kritis-config.yaml`
You can see what each field mean [here](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition)

2. Once you have all these defined, and code generated, you can test it on your cluster
a.  Deploy your Custom Resource using
```
kubectl apply -f kritis-config.yaml
```
b. See if you can list your newly created custom resource
```
kubectl get <resource_plural_name>
```
The resource_plural_name is mentioned in the CRD config. 
For `BuildPolicy` it is `buildpolicies ` and is mentioned [here](https://github.com/grafeas/kritis/blob/493cd0d4f73bfc6aa8804951c1e6cc8da380c336/artifacts/build-policy-crd.yaml#L10)
c. Once you have created a resource definition, you can now create resources of type `KritisConfig` and deploy it using `kubectl apply`
Look here for examples [here](https://github.com/grafeas/kritis/tree/493cd0d4f73bfc6aa8804951c1e6cc8da380c336/artifacts/examples) .
d . if you run `kubectl get` again, you shd be able to see this new resource.


[...]

PR #299, opened 2018-10-29
Added KritisConfig CRD to Kritis

aysylu:


aysylu:
The first PR to address #224

aysylu:
I confirmed that this works locally by creating a KritisConfig object, deploying the custom resource, and listing it successfully:

```
$ vim kritis-config.yaml
kind: KritisConfig
metadata:
  name: my-config
  namespace: default
$ kubectl apply -f kritis-config.yaml 
kritisconfig.kritis.grafeas.io/my-config created
$ kubectl get kritisconfigs
NAME        AGE
my-config   8s
```

PR #301, opened 2018-10-30
Added KritisConfig CRD to Kritis

aysylu:
The first PR to address #224.

I confirmed that this works locally by creating a KritisConfig object, deploying the custom resource, and listing it successfully:

```
$ vim kritis-config.yaml
kind: KritisConfig
metadata:
  name: my-config
  namespace: default
$ kubectl apply -f kritis-config.yaml 
kritisconfig.kritis.grafeas.io/my-config created
$ kubectl get kritisconfigs
NAME        AGE
my-config   8s
```


tejal29:
thanks! will take a look in pm

pkg/kritis/apis/kritis/v1beta1/types.go
```diff
@@ -116,3 +116,22 @@ type AttestationAuthorityList struct {
...
 	Items []AttestationAuthority `json:"items"`
 }
+
+// +genclient
+// +genclient:noStatus
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type KritisConfig struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
```
tejal29: Right now, the Kritis Config has no Spec. Are planning to add it late?

aysylu:
Thanks for the review!

PR #303, opened 2018-11-01
Extracted configurable values into KritisConfig

aysylu:
Specifically, this covers the following configuration values:
 
* storage backend for metadata storage and retrieval,
* cron job interval,
* the address of the Kritis server.

This lays the groundwork for making Kritis more configurable (#224).


cmd/kritis/admission/main.go
```diff
@@ -65,8 +56,37 @@ func main() {
...
+	if len(kritisConfigs) == 0 {
+		glog.Errorf("No KritisConfigs found in any namespace")
+		return
+	} else if len(kritisConfigs) > 1 {
+		glog.Errorf("More than 1 KritisConfig found, will use the 0th object")
+	}
+
+	kritisConfig := kritisConfigs[0]
+	// TODO(https://github.com/grafeas/kritis/issues/304): Use CRD validation instead
+	if len(kritisConfig.Spec.MetadataBackend) == 0 {
```
tejal29: since these are strings you check for empty string 
```if kritisConfig.Spec.MetadataBackend != "" {
} else {
}
```

aysylu: SG, PTAL.

integration/testdata/kritis-server/kritis-config.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: kritis.grafeas.io/v1beta1
+kind: KritisConfig
+metadata:
+  name: my-kritis-config
+  namespace: default
```
priyawadhwa: I think this may be the bug -- the file specifies the namespace as "default" but the integration tests try to deploy to a different namespace each time

integration/run_test.go
```diff
@@ -287,6 +287,12 @@ func setUp(t *testing.T) (kubernetes.Interface, *v1.Namespace, func(t *testing.T
 	if out, err := integration_util.RunCmdOut(cmd); err != nil {
 		t.Fatalf("testing error: %v\nout: %s", err, out)
 	}
+
+	cmd = exec.Command("kubectl", "apply", "-f", "kritis-server/kritis-config.yaml")
```
priyawadhwa: I think the path here needs to be `testdata/kritis-server/kritis-config.yaml`

integration/testdata/kritis-server/kritis-config.yaml
aysylu: From the discussion offline, several of the namespace-related issues were fixed. Thanks!

integration/run_test.go
aysylu: Good catch, thanks!

aysylu:
Currently, there's 1 test failure due to an issue (most likely) unrelated to this PR. PR #308 addresses this issue, so I'll wait to re-run the tests until after it's been merged.

cmd/kritis/admission/main.go
```diff
@@ -65,8 +56,37 @@ func main() {
 		fmt.Println(version.Commit)
 		return
 	}
+
+	kritisConfigs, err := kritisconfig.KritisConfigs("default")
```
priyawadhwa: This assumes users will deploy the config to the "default" namespace, so you might want to change it to `""` to get all configs in all namespaces!

cmd/kritis/admission/main.go
```diff
@@ -65,8 +56,37 @@ func main() {
...
+	if err != nil {
+		errMsg := fmt.Sprintf("error getting kritis config: %v", err)
+		glog.Errorf(errMsg)
+		return
+	}
+	if len(kritisConfigs) == 0 {
+		glog.Errorf("No KritisConfigs found in any namespace")
+		return
```
priyawadhwa: If the user doesn't deploy a KritisConfig, should kritis assume defaults? Or is it required that the user also deploys a KritisConfig?

cmd/kritis/admission/main.go
```diff
@@ -65,8 +56,37 @@ func main() {
...
+	}
+	if kritisConf.Spec.CronInterval == "" {
+		glog.Errorf("No KritisConfigs CronInterval is defined in the spec")
+		return
+	}
+	if kritisConf.Spec.ServerAddr == "" {
+		glog.Errorf("No KritisConfigs ServerAddr is defined in the spec")
+		return
+	}
```
priyawadhwa: Similar to the above comment, here we could assume defaults if nothing is provided?

cmd/kritis/admission/main.go
aysylu: Good point, this is a leftover from debugging. Fixed.

cmd/kritis/admission/main.go
aysylu: Good point, I think it should. For now, there's no need to require that the user deploys a KritisConfig. Updated the logic to reflect this.

cmd/kritis/admission/main.go
aysylu: Updated the logic to use the defaults.

cmd/kritis/admission/main.go
```diff
@@ -65,8 +65,39 @@ func main() {
...
+
+	metadataBackend := DefaultMetadataBackend
+	cronInterval := DefaultCronInterval
+	serverAddr := DefaultServerAddr
+
+	if len(kritisConfigs) == 0 {
+		glog.Errorf("No KritisConfigs found in any namespace, will assume the defaults")
+	} else {
+		if len(kritisConfigs) > 1 {
+			glog.Errorf("More than 1 KritisConfig found, will use the 0th object")
```
priyawadhwa: WDYT of returning an error here instead to enforce one KritisConfig/cluster?

aysylu: Good question. I've considered this and at the time decided that it was better to parse any config than failing the program at this point. However, given that the CRD is per-cluster, I can see the value in the argument that this is an unexpected condition. Updated the condition to return an error, open to feedback from others if there are valid cases to have more than 1 KritisConfig defined.

PR #307, opened 2018-11-02
Added copying of artifacts folder with CRDs

aysylu:
This will allow us to remove redundant code in https://github.com/grafeas/kritis/blob/master/helm-hooks/preinstall/crd.go that redefines the CRDs. Does a cleanup for #224.

### DIFF
--- a/helm-hooks/Dockerfile
+++ b/helm-hooks/Dockerfile
@@ -26,4 +26,7 @@ ENV KUBECTL_VERSION v1.10.0
 RUN curl -Lo /usr/local/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl  && \
     chmod +x /usr/local/bin/kubectl
 
-COPY --from=0 /go/src/github.com/grafeas/kritis/out/${stage}  /${stage} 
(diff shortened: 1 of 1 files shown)


helm-hooks/Dockerfile
```diff
@@ -26,4 +26,7 @@ ENV KUBECTL_VERSION v1.10.0
 RUN curl -Lo /usr/local/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl  && \
     chmod +x /usr/local/bin/kubectl
 
-COPY --from=0 /go/src/github.com/grafeas/kritis/out/${stage}  /${stage} 
+COPY --from=0 /go/src/github.com/grafeas/kritis/out/${stage}  /${stage}
+
+# Copy artifacts to the image.
+COPY --from=0 /go/src/github.com/grafeas/kritis/artifacts /${stage}/artifacts
```
priyawadhwa: This looks good, except I'd probably copy it into just `/artifacts`, otherwise the three different images will store the artifacts directory in different locations (`/predelete/artifacts` vs `/preinstall/artifacts` vs `/postinstall/artifacts`)

aysylu: I see what you mean, thanks! Updated, PTAL.

PR #308, opened 2018-11-06
Fixed TLS secret generation

aysylu:
It now waits for the CSR to be first fully generated, before creating the secret.

### DIFF
--- a/helm-hooks/preinstall/preinstall.go
+++ b/helm-hooks/preinstall/preinstall.go
@@ -171,7 +171,7 @@ func createTLSSecret() {
 	}
 	certStr = strings.TrimPrefix(certStr, "'")
 	certStr = strings.TrimSuffix(certStr, "'")
-	cert = []byte(certStr)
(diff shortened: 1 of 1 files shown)


priyawadhwa:
This looks great, and the tests passed! The only suggestion I have is to take the logic out into a `waitForCSR()` function, similar to how we wait for the preinstall pod [here](https://github.com/grafeas/kritis/blob/master/helm-hooks/postinstall/postinstall.go#L41) so that it matches the rest of the code and there isn't any risk of getting stuck in an infinite loop.

aysylu:
> This looks great, and the tests passed! The only suggestion I have is to take the logic out into a `waitForCSR()` function, similar to how we wait for the preinstall pod [here](https://github.com/grafeas/kritis/blob/master/helm-hooks/postinstall/postinstall.go#L41) so that it matches the rest of the code and there isn't any risk of getting stuck in an infinite loop.

Good idea. Done.

PR #312, opened 2018-11-27
Fix admission control ignores configured server address

liron-l:


aysylu:
Thanks for the fix!

aysylu:
Waiting for the kokoro tests to complete running, then will merge the PR on successful completion.

PR #313, opened 2018-11-28
Fix reading attestation key

liron-l:
From [docuemnation](https://github.com/grafeas/kritis/blob/master/docs/tutorial.md) it appears that the PublicKeyData is base64 encoded, thus cannot be used for `NewPgpKey`.
```
PUBLIC_KEY=`base64 gpg.pub -w 0`
publicKeyData: $PUBLIC_KEY
```

Also added UT for `NewPgpKey`

### DIFF
--- a/pkg/kritis/review/review.go
+++ b/pkg/kritis/review/review.go
@@ -17,6 +17,7 @@ limitations under the License.
 package review
 
 import (
+	"encoding/base64"
 	"fmt"
 
 	"github.com/golang/glog"
@@ -126,12 +127,12 @@ Please see instructions `, image)
 	}
 	keys := map[string]string{}
 	for _, auth := range auths {
-		s, err := attestation.NewPgpKey("", auth.Spec.PublicKeyData)
+		key, fingerprint, err := fingerprint(auth.Spec.PublicKeyData)
 		if err != nil {
 			glog.Errorf("Error parsing key for %q: %v", auth.Name, err)
 			continue
 		}
-		keys[s.Fingerprint()] = auth.Spec.PublicKeyData
+		keys[fingerprint] = key
 	}
 	for _, a := range attestations {
 		if err = host.VerifyAttestationSignature(keys[a.KeyID], a.Signature); err != nil {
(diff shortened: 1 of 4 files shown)


aysylu:
Hi @liron-l! I'm reviewing this PR and it appears that the Travis test is currently failing. Also, how did you discover this issue? Following the tutorial worked for me, so I'm wondering what aspect is broken here.

pkg/kritis/review/review.go
```diff
@@ -170,12 +171,12 @@ func (r Reviewer) addAttestations(image string, atts []metadata.PGPAttestation,
...
 	keys := map[string]string{}
 	for _, auth := range auths {
-		s, err := attestation.NewPgpKey("", auth.Spec.PublicKeyData)
+		key, fingerprint, err := fingerprint(auth.Spec.PublicKeyData)
 		if err != nil {
 			glog.Errorf("Error parsing key for %q: %v", auth.Name, err)
 			continue
 		}
-		keys[auth.Name] = s.Fingerprint()
+		keys[fingerprint] = key
```
aysylu: I think this is changing the existing behavior? It should be `keys[auth.Name] = fingerprint`

liron-l: Thanks sorry for that.

liron-l:
@priyawadhwa i think this was caused by a regression from 34757469f26cd2848857d7db064efc1bfed13729. Specifically:
1. The k8s secrets are loaded as standard text files (so calling `NewPgpKey` with plain secrets works).
2. The `publicKeyData` is loaded as the base64 encoded public key
3. In 34757469f26cd2848857d7db064efc1bfed13729 the code that decodeds the base64 string was removed from `NewPgpKey`; so calling `NewPgpKey('base64string',..)` appears to be broken.
4. `NewPgpKey` does not accept base64 strings (`armor.Decode(r)` assumes plain text) and k8s secrets are not loaded as base64 strings, so i fixed the comment

I can also patch it such that `NewPgpKey` will automatically detect whether the string is plain or base64 encoded, but I thought the proposed change is a little more robust.

But maybe i'm missing something.

priyawadhwa:
Thanks for explaining! So to clarify, #211 assumes that the k8s api fetches the decoded secret, but that actually isn't the case?

liron-l:
@priyawadhwa there are two different flows, one that reads the k8s secrets (which are not base64 encoded) and one that reads the AttestationAuthority public key data (which is base64 encoded).  This change just decodes the AttestationAuthority public key data before creating the pgp struct (technically, it finalizes #211)

priyawadhwa:
Ah gotcha, thanks!

PR #314, opened 2018-12-05
Add AttestationAuthoriyNames to ImageSecurityPolicy Spec.

tejal29:
This is to differentiate between Attestation Authorities configured for ImageSecurityPolicy and other Policies like BuildPolicy.
BuildPolicy already has a [AttestationAuthorityName](https://github.com/grafeas/kritis/blob/master/pkg/kritis/apis/kritis/v1beta1/types.go#L72)

I added a List of names here instead of single value because in case one secret is compromised, we can start attesting using a new secret.
The old attestation can still be verified using the old secret.
Once all images have been attested using new secret, administrator can go and deleted the compromised secret.


pkg/kritis/apis/kritis/v1beta1/securitypolicy.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +genclient:noStatus
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type ImageSecurityPolicy struct {
```
priyawadhwa: Are these being added again because we're switching the config to v1beta1?

pkg/kritis/review/review_test.go
```diff
@@ -370,3 +374,74 @@ func makeAtt(ids []string) []metadata.PGPAttestation {
...
+				Spec: v1beta1.ImageSecurityPolicySpec{
+					AttestationAuthorityNames: tc.aList,
+				},
+			}
+			auths, err := r.getAttestationAuthoritiesForISP(isp)
+			if (err != nil) != tc.shdErr {
+				t.Fatalf("expected review to return error %t, actual error %s", tc.shdErr, err)
+			}
+			if len(auths) != tc.expectedLen {
+				t.Fatalf("expected review to return error %t, actual error %s", tc.shdErr, err)
```
tstromberg: Make this message more accurate. Also, use t.Errorf when possible in preference to t.Fatalf. They both behave the same in this example, however.

PR #317, opened 2018-12-27
Set the BAZEL_VERSION env var

aysylu:
This fixes the integration tests failure with the error message: `common.sh: line 5: BAZEL_VERSION: unbound variable`.


aysylu:
Test expected to pass in the subsequent runs after the merge. If it doesn't pass for #316, I'll revert or fix this in another PR.

PR #325, opened 2019-02-11
fix yaml syntax

oke-py:
`whitelistCVEs` is a slice of string, and so it needs a heading '-' in yaml like [this](https://github.com/grafeas/kritis/blob/master/integration/testdata/image-security-policy/my-isp.yaml#L14-L16).


aysylu:
Thanks for catching and fixing this!

aysylu:
The kokoro test failure is unrelated to this PR, and I'm investigating this separately. Thanks for your contribution, @oke-py!

PR #327, opened 2019-02-12
testing kokoro

aysylu:


aysylu:
The tests are now passing, likely due to the fix in #324. Will re-open investigation if subsequent PRs are having a problem, for now unable to reproduce the issue after merging the fix.

PR #332, opened 2019-02-25
Skip reviewing if there is no ImageSecurityPolicy in the namespace

dragon3:


dragon3:
Sorry again...
I've created this pull request by mistake...

aysylu:
No problem @dragon3! We actually plan to provide default fallback for when policies are undefined for `0.1.0`, so this is in-line with our goal in spirit.:)

PR #334, opened 2019-03-01
Adding BuildPolicy Verification on to Kritis Flow

tejal29:
In this commit,
1. AdmissionConfig declares BuildPolicy Reviewer.
2. The BuildPolicyReview is right now the google cloud build policy
reviewer. This should be replaced by a generic reviewer.
3. Admission flow now reviews BuildPolicy along with ISP
4. The review process for Build Policy is
   a. Fetch all the BuildPolicies in the namespace
   b. Convert container image to google cloud build Provenance struct.
      This needs to go away in future.
   c. Call the policy ValidateAndSign method.
5. ValidateAndSign will first
  - check if valid attestation exists for this container image.
     if it does, it will return immediately which will suffice the use
     case we have now.
  - if no valid attestation is found, it will verify the source of build
  and either deny or allow.

Please remember this is not tested and might need some work.

### DIFF
--- a/pkg/kritis/admission/admission.go
+++ b/pkg/kritis/admission/admission.go
@@ -21,9 +21,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-
-	"github.com/grafeas/kritis/pkg/kritis/metadata/containeranalysis"
-	"github.com/grafeas/kritis/pkg/kritis/metadata/grafeas"
+	"strings"
 
 	"github.com/golang/glog"
 	"github.com/grafeas/kritis/cmd/kritis/version"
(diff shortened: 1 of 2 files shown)


tejal29:
hi, 
@dragon3 the code actually creates a BuildProvenance Object from your Build Policy. 
`BuildPolicy` CRD looks like this
```
kind: BuildPolicy
metadata:
  name: kritis-bp
  namespace: default
spec:
  attestationAuthorityName: kritis-authority
  buildRequirements:
    builtFrom: https://source.developers.google.com/p/binauthz-demo1-builder/r/demo-app:build.*
```

In your case you can ignore `builtFrom` and use the attestationAuthorityName to point to K8 secret which you use to sign your image?
Does that make sense?

pkg/kritis/admission/admission.go
```diff
@@ -229,13 +235,56 @@ func createDeniedResponse(ar *v1beta1.AdmissionReview, message string) {
 
 func reviewImages(images []string, ns string, pod *v1.Pod, ar *v1beta1.AdmissionReview, config *Config) {
 	// NOTE: pod may be nil if we are reviewing images for a replica set.
-	glog.Infof("Reviewing images for %s in namespace %s: %s", pod, ns, images)
+	deny := false
+	errs := []string{}
+	glog.Infof("Reviewing Build Policy for images for %s in namespace %s: %s", pod, ns, images)
+	if err := reviewBuildPolicy(images, ns, config); err != nil {
+		deny = true
+		errs = append(errs, err.Error())
```
tejal29: should we fail fast here?

tejal29:
Discussing with @aysylu  and we decided with adding a new GenericAttestationPolicy which will only contain the `Attestation authority Spec name`

aysylu:
Per @tejal29's comment above, this is now superseded by #344.

Issue #343, opened 2019-04-23
Go linter should give suggestions compatible with Google-internal style


aysylu:
**Is your feature request related to a problem? Please describe.**
Currently, `golangci-lint` gives suggestions that contradict Google-internal style. This is an issue mainly because it causes a divergence in adopted styles across Kritis and Grafeas.

**Describe the solution you'd like**
Adopt a linter that streamlines the styles across the 2 codebases.

**Describe alternatives you've considered**
Keep the codebases adhering to different styles enforced by linters.

PR #344, opened 2019-04-24
GenericAttestationPolicy is now enforced in Kritis

aysylu:
Supersedes #334.

### DIFF
--- a/pkg/kritis/admission/admission.go
+++ b/pkg/kritis/admission/admission.go
@@ -368,7 +368,7 @@ func getReviewer(client metadata.Fetcher) reviewer {
 	})
 }
 
-// reviewer interface defines Kritis Reviewer struct.
+// reviewer interface defines Kritis Reviewer struct, useful for mocking in tests
 type reviewer interface {
 	ReviewGAP(images []string, isps []kritisv1beta1.GenericAttestationPolicy, pod *v1.Pod) error
 	ReviewISP(images []string, isps []kritisv1beta1.ImageSecurityPolicy, pod *v1.Pod) error
--- a/pkg/kritis/admission/admission.go
+++ b/pkg/kritis/admission/admission.go
@@ -31,6 +31,7 @@ import (
 	kritisv1beta1 "github.com/grafeas/kritis/pkg/kritis/apis/kritis/v1beta1"
 	kritisconstants "github.com/grafeas/kritis/pkg/kritis/constants"
 	"github.com/grafeas/kritis/pkg/kritis/crd/authority"
+	"github.com/grafeas/kritis/pkg/kritis/crd/genericattestation"
 	"github.com/grafeas/kritis/pkg/kritis/crd/securitypolicy"
 	"github.com/grafeas/kritis/pkg/kritis/metadata"
 	"github.com/grafeas/kritis/pkg/kritis/review"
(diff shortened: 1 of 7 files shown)


pkg/kritis/admission/admission.go
```diff
@@ -31,35 +31,38 @@ import (
...
+	"github.com/grafeas/kritis/pkg/kritis/crd/genericattestation"
 	"github.com/grafeas/kritis/pkg/kritis/crd/securitypolicy"
 	"github.com/grafeas/kritis/pkg/kritis/metadata"
 	"github.com/grafeas/kritis/pkg/kritis/review"
 	"github.com/grafeas/kritis/pkg/kritis/secrets"
 	"github.com/grafeas/kritis/pkg/kritis/violation"
 	"k8s.io/api/admission/v1beta1"
 	appsv1 "k8s.io/api/apps/v1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
clrprod: This shouldn't be needed? The name of import is always the right most element (which is v1).

pkg/kritis/admission/admission.go
```diff
@@ -230,6 +233,18 @@ func createDeniedResponse(ar *v1beta1.AdmissionReview, message string) {
 func reviewImages(images []string, ns string, pod *v1.Pod, ar *v1beta1.AdmissionReview, config *Config) {
 	// NOTE: pod may be nil if we are reviewing images for a replica set.
 	glog.Infof("Reviewing images for %s in namespace %s: %s", pod, ns, images)
+	client, err := admissionConfig.fetchMetadataClient(config)
+	if err != nil {
+		errMsg := fmt.Sprintf("error getting metadata client: %v", err)
+		glog.Errorf(errMsg)
+		createDeniedResponse(ar, errMsg)
```
clrprod: It's clearer if function output is a returned parameter from a function - it looks like ar may only be return, could it be moved over?

i.e.
func reviewImages(...) *v1beta1.AdmissionReview

pkg/kritis/admission/admission.go
```diff
@@ -339,8 +368,8 @@ func getReviewer(client metadata.Fetcher) reviewer {
 	})
 }
 
-// reviewer interface defines an Kritis Reviewer Struct.
-// TODO: This will be removed in future refactoring.
+// reviewer interface defines Kritis Reviewer struct.
 type reviewer interface {
```
clrprod: Does this interface need to exist? Can the underlying type just be used instead?

pkg/kritis/review/review_test.go
```diff
@@ -111,7 +111,134 @@ func TestHasValidAttestations(t *testing.T) {
...
+	}
+	validAtts := []metadata.PGPAttestation{{Signature: sig, KeyID: secFpr}}
+
+	invalidSig, err := util.CreateAttestationSignature(testutil.IntTestImage, sec)
+	if err != nil {
+		t.Fatalf("unexpected error %v", err)
+	}
+	invalidAtts := []metadata.PGPAttestation{{Signature: invalidSig, KeyID: secFpr}}
+
+	var gap = v1beta1.GenericAttestationPolicy{
```
clrprod: This is equivalent to gap :=

pkg/kritis/admission/admission_test.go
```diff
@@ -105,6 +105,9 @@ func Test_AdmissionResponse(t *testing.T) {
 		{"valid response when no review error", false, constants.SuccessStatus, constants.SuccessMessage},
 		{"valid response when review error", true, constants.FailureStatus, fmt.Sprintf("found violations in %s", testutil.QualifiedImage)},
 	}
+	mockGAP := func(namespace string) ([]kritisv1beta1.GenericAttestationPolicy, error) {
```
clrprod: This mock shouldn't be needed - can't grafeas be imported and spun up in memory only mode?

pkg/kritis/admission/admission.go
aysylu: This is due to the `golangci-lint` style. I filed #343 to straighten out this story. In the meantime, I'd like to have fewer lint comments for now, since they still catch useful issues.

pkg/kritis/review/review_test.go
aysylu: Good point, updated here and in the ReviewISP test.

pkg/kritis/admission/admission_test.go
aysylu: I'm not sure I understand the suggestion. This is providing the Generic Attestation Policy, which is a Kritis-specific thing, unrelated to Grafeas.

clrprod: Ah got it - I thought it was stored in grafeas for some reason.

pkg/kritis/admission/admission.go
aysylu: The interface is useful for mocking it in the admission_test. I added a comment and enhanced the tests for this.

pkg/kritis/admission/admission.go
aysylu: That's a good point. Given the style of passing this as pointer receivers, I'll keep this as is for now, and consider a larger refactoring later.

tejal29:
Awesome!!!

PR #352, opened 2019-06-19
Validated attestation

nenaddedic:
The main focus of this PR is to define two interfaces to work with attestations:
- ValidatedAttestation which represents a trusted, verified attestation
- ValidatingTransport which allows the caller to obtain ValidatedAttestations for a given image

ValidatedAttestation can contain arbitrary JSON payload, and can therefore represent a wide variety of verified metadata.  ValidatedAttestation is deliberately stripped of signatures, because the producer of ValidatedAttestation must establish that the payload is trusted by verifying the signature.

Additionally, this PR changes the ISP and GAP review routines to use ValidatedAttestation.


pkg/kritis/attestation/validated_attestation.go
```diff
@@ -0,0 +1,26 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package attestation
+
+// ValidatedAttestation represents a trusted and verified attestation made by
+// named attestor about named artifact.  Any claims are contained in the
```
aysylu: Rename this to AttestationAuthority for consistency with the current state of the codebase?

aysylu: I'm not familiar with this term, what does it mean?

aysylu: It'd be helpful to provide examples of what you mean by artifact here, in the form "(e.g. ...)", for the reader.

pkg/kritis/review/validating_transport.go
```diff
@@ -0,0 +1,71 @@
...
+	GetValidatedAttestations(artifact string) ([]attestation.ValidatedAttestation, error)
+}
+
+// Implements ValidatingTransport.
+type AttestorValidatingTransport struct {
+	Client   metadata.Fetcher
+	Attestor v1beta1.AttestationAuthority
+}
+
+func (avt AttestorValidatingTransport) GetValidatedAttestations(artifact string) ([]attestation.ValidatedAttestation, error) {
```
aysylu: change to `*AttestorValidatingTransport`?

aysylu: Because Kritis is all about container images, let's rename `artifact` to `image` for consistency and clarity.

pkg/kritis/attestation/validated_attestation.go
```diff
@@ -0,0 +1,26 @@
...
+// payload, which is a map representation of a JSON object.
+type ValidatedAttestation struct {
+	AttestorName string
+	ArtifactName string
```
aysylu: s/ArtifactName/Image?

pkg/kritis/review/validating_transport.go
```diff
@@ -0,0 +1,71 @@
...
+		glog.Errorf("Error parsing key for %q: %v", avt.Attestor.Name, err)
+		return nil, err
+	}
+	keys[fingerprint] = key
+
+	out := []attestation.ValidatedAttestation{}
+	host, err := container.NewAtomicContainerSig(artifact, map[string]string{})
+	if err != nil {
+		glog.Error(err)
+		// TODO: maybe do partial validation for other images
```
aysylu: I'm not sure I follow what this TODO is proposing, could you please clarify? ditto for below

pkg/kritis/review/validating_transport.go
```diff
@@ -0,0 +1,71 @@
...
+	attestations, err := avt.Client.Attestations(artifact)
+	if err != nil {
+		glog.Error(err)
+		// TODO: maybe do partial validation for other images
+		return nil, err
+	}
+	for _, a := range attestations {
+		if err = host.VerifyAttestationSignature(keys[a.KeyID], a.Signature); err != nil {
+			glog.Errorf("Could not find or verify attestation for attestor %s", a.KeyID)
+			glog.Errorf(err.Error())
```
aysylu: nit: merge this into the error line above with %v?

pkg/kritis/review/validating_transport.go
```diff
@@ -0,0 +1,71 @@
...
+		} else {
+			out = append(out, attestation.ValidatedAttestation{AttestorName: avt.Attestor.Name, ArtifactName: artifact})
```
aysylu: Would the `Payload` field be set here or somewhere else? I'm trying to understand how it comes into play here.

pkg/kritis/review/validating_transport_test.go
```diff
@@ -0,0 +1,102 @@
...
+	anotherSig, err := util.CreateAttestationSignature(testutil.IntTestImage, successSec)
+	if err != nil {
+		t.Fatalf("unexpected error %v", err)
+	}
+	tcs := []struct {
+		name         string
+		expected     []attestation.ValidatedAttestation
+		attestations []metadata.PGPAttestation
+	}{
+		{"at least one valid sig", []attestation.ValidatedAttestation{
```
aysylu: nit: could we provide names of fields for readability here? eg `name` and `expected` and `attestations`

pkg/kritis/review/validating_transport_test.go
```diff
@@ -0,0 +1,102 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/grafeas/kritis/pkg/kritis/apis/kritis/v1beta1"
+	"github.com/grafeas/kritis/pkg/kritis/attestation"
+	"github.com/grafeas/kritis/pkg/kritis/metadata"
+	"github.com/grafeas/kritis/pkg/kritis/testutil"
+	"github.com/grafeas/kritis/pkg/kritis/util"
+)
+
+func TestValidatingTransport(t *testing.T) {
```
aysylu: Let's split success tests from error tests here.

pkg/kritis/review/review.go
```diff
@@ -72,12 +70,22 @@ func (r Reviewer) ReviewGAP(images []string, gaps []v1beta1.GenericAttestationPo
...
 			}
-			isAuthAttested, _ := r.fetchAndVerifyAttestations(image, auths, pod)
-			if isAuthAttested {
-				imgAttested = true
-				break
+			for _, auth := range auths {
+				transport := AttestorValidatingTransport{Client: r.client, Attestor: auth}
+				attestations, err := transport.GetValidatedAttestations(image)
+				if err != nil {
+					// TODO or something better
```
aysylu: let's at least log the error for now

pkg/kritis/review/review.go
```diff
@@ -72,12 +70,22 @@ func (r Reviewer) ReviewGAP(images []string, gaps []v1beta1.GenericAttestationPo
 			if err != nil {
 				return err
 			}
-			isAuthAttested, _ := r.fetchAndVerifyAttestations(image, auths, pod)
-			if isAuthAttested {
-				imgAttested = true
-				break
+			for _, auth := range auths {
+				transport := AttestorValidatingTransport{Client: r.client, Attestor: auth}
```
aysylu: extract into its own method, for readability?

pkg/kritis/attestation/validated_attestation.go
```diff
@@ -0,0 +1,26 @@
...
+	Payload      map[string]interface{}
```
aysylu: From the current state of the PR, I'm not sure how `Payload` is used. Is this an omission or is there another PR to follow? Either way, I'd love to understand the goal of this field better.

pkg/kritis/review/validating_transport.go
nenaddedic: Nothing to clarify, who knows why I left this here.  Removed.

pkg/kritis/review/validating_transport.go
nenaddedic: Yes, most likely here.  For example when we use build provenance in policy checks, this would be about where you populate the payload with the build provenance details.

pkg/kritis/attestation/validated_attestation.go
nenaddedic: It is deliberate, but unused.

When we start supporting attestations with payload (again, build provenance example), this is where I expect the payload to go.  I think it is good to keep it as a reminder of sorts, even if we don't use it right away.  WDYT?

In the near-ish future, I expect a ValidatedAttestation like:
{
  AttestorName: "some-attestor",
  Image: "gcr.io/image@sha256:abcd"
  Payload: { "github_repo": "yadayada" }
}

pkg/kritis/attestation/validated_attestation.go
nenaddedic: I rewrote the comment to say "container image" instead, do you think an example is still needed?

nenaddedic: I rewrote the comment a bit more to motivate the payload field, clarify it, and juxtapose with a reasonably common technology (JWT).  Does this help?

pkg/kritis/review/review.go
nenaddedic: What should I extract?  The loop that goes over auths?  Something else?

pkg/kritis/review/validating_transport_test.go
nenaddedic: I skimmed this from TestHasValidAttestations (review_test.go) and as far as I can tell there are no error tests here.  So I don't think there's anything to split.  Do you want me to add error tests?

pkg/kritis/attestation/validated_attestation.go
aysylu: Let's remove it for now and add it when we have explicit plans to use it. Otherwise, it's creating more confusion at this point. Also, the name `payload` is so generic that we might want to revisit this in the future.

pkg/kritis/review/review.go
aysylu: Yes, the loop that goes over auths please, for readability here and in the method below.

pkg/kritis/review/validating_transport_test.go
aysylu: I initially read the tests similar to `no valid sig` as error tests, but you're right. Yes please, let's add error tests here, this is generally the pattern we'd like to follow in our codebase (and we're not there yet).

nenaddedic: Done.

pkg/kritis/review/review.go
nenaddedic: Done.

pkg/kritis/attestation/validated_attestation.go
nenaddedic: Done.

pkg/kritis/attestation/validated_attestation.go
```diff
@@ -0,0 +1,29 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package attestation
+
+// ValidatedAttestation represents a trusted and verified attestation made by
+// the named attestation authority about the named container image.
+// An example ValidateAttestation may look like:
```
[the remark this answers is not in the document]
nenaddedic: Done.

aysylu:
Thanks for your contribution, @nenaddedic!

PR #354, opened 2019-06-28
Refactored logic for clarity of image admission process

aysylu:
Fixed severity in logs when no policies are found.


aysylu:
The previously failing integration test was a flake.

PR #355, opened 2019-07-03
Updated Kritis server to pull certificates for Grafeas from secrets

aysylu:


aysylu:
> Is there any way to write a test which spins up grafeas in minikube and makes sure it starts cleanly? That would probably catch lots of silly issues here.

I filed #357 to add this test. It'd be very similar to existing ones, but will first try to install Grafeas helm chart, before installing Kritis.

pkg/kritis/metadata/grafeas/config.go
```diff
@@ -0,0 +1,63 @@
...
+	CAFile   string `yaml:"cafile"`   // A PEM encoded CA's certificate file
+}
+
+// defaultConfig is a configuration that can be used as a fallback value.
+func defaultConfig() *CertConfig {
+	return &CertConfig{
+		CertFile: "",
+		KeyFile:  "",
+		CAFile:   "",
+	}
```
clrprod: FYI - this should be equivalent to return &CertConfig{} (and in fact you may not need the default function at all).

aysylu: Good point, done.

PR #360, opened 2019-07-16
Fixed assumptions for KritisConfig's Grafeas address

aysylu:
1. `ServerName` in `tls.Config` is now set to Grafeas address, which is the `kubectl svc` name;
2. Removed expectation that the Grafeas address is specified in the `host:port` format.


aysylu:
> Is this just a optimization? Or is it worth writing a test that would have broken with the old form?

The current tests broke, which is good. Looking further into this, I no longer believe this is the correct approach, as `grpc.Dial()` expects `host:port` format, so that's what we should be specifying in KritisConfig. Withdrawing this PR to pursue another approach.

PR #365, opened 2019-07-23
Added global allowlist logic to GenericAttestationPolicy

aysylu:
This enables for Grafeas server restarts when Kritis is already deployed and GenericAttestationPolicy is set.


aysylu:
Integration test failure is another flake :/ The tests passed locally.

PR #375, opened 2019-08-07
Make setup_helm.sh executable

Katee:
Found while following https://github.com/grafeas/kritis/blob/master/docs/standalone/README.md#installation-steps


aysylu:
The commit shows zero diffs, am I missing something?

Katee:
It changes the mode of the file from `100644 → 100755`. This means that the line `./setup_helm.sh` from https://github.com/grafeas/kritis/tree/master/docs/standalone#installation-steps will work because that file is now executable.

aysylu:
> It changes the mode of the file from `100644 → 100755`. This means that the line `./setup_helm.sh` from https://github.com/grafeas/kritis/tree/master/docs/standalone#installation-steps will work because that file is now executable.

I see now, thanks!

Issue #376, opened 2019-08-08
Passphrase in k8s secrets fails base64 check


ooq:
## Expected Behavior
Secrets with passphrase can be successfully fetched.

## Actual Behavior
`secrets.Fetch()` fails with `base64 decode failed illegal base64 data at input byte 8`.

## Steps to Reproduce the Problem

1. Set PHRASE environment variable:
```
PHRASE=passphrase1
```
2. Create public and private key pair with the passphrase1 as passphrase:
```
gpg --quick-generate-key --yes attestor@example.com
gpg --armor --export attestor@example.com > gpg.pub
gpg --armor --export-secret-keys attestor@example.com > gpg.priv
```
3. Create k8s secret in default namespace
```
kubectl create secret generic attestor --from-file=public=gpg.pub --from-file=private=gpg.priv --from-literal=passphrase=${PHRASE}
```
4. `go run` the following program:
```
package main

import (
"log"

"github.com/grafeas/kritis/pkg/kritis/secrets"
)

func main() {
    s, err := secrets.Fetch("default", "attestor")
    if err != nil {
    log.Fatal(err)
    }
    log.Printf("Retrieved secret for 'attestor': %v", s)
}
```
## Fix Proposal
This bug is caused by base64 decoding on the fetched passphrase in `secret.Fetch()`. This base64 is redundant since the fetched passphrase has been decoded. Propose to remove the decoding.

ooq:
Fixed by #379. Closing.

PR #379, opened 2019-08-08
[ISSUE-376] Remove base64 decoding for passphrase.

ooq:
Fixing #376 by remove base64 decoding for passphrase.


ooq:
For sure!

ooq:
Tests added.

pkg/kritis/secrets/secrets_test.go
```diff
@@ -31,56 +31,99 @@ import (
...
 	name       string
 	secretName string
 	shdErr     bool
 	expected   *PGPSigningSecret
 }{
-	{"good", "good-sec", false, &PGPSigningSecret{SecretName: "good-sec", PgpKey: pgpKey}},
-	{"bad1", "bad1-sec", true, nil},
-	{"bad2", "bad2-sec", true, nil},
-	{"notfound", "not-present", true, nil},
+	{"good-nopass", "good-sec-nopass", false, &PGPSigningSecret{SecretName: "good-sec-nopass", PgpKey: pgpKey}},
```
aysylu: consider removing the `secretName` field and just re-using the `name` in the tests, since they're so similar

pkg/kritis/secrets/secrets_test.go
```diff
@@ -31,56 +31,99 @@ import (
...
 			}
 		})
 	}
 }
 
+// Custom methods to check if two secrets are equal.
+// We avoid using reflect.DeepEqual() directly bacause a passphrase-protected
+// packet.PrivateKey can contain a function field.
+func isSecretEqual(x, y *PGPSigningSecret) bool {
+	if (x == nil || y == nil) {
```
aysylu: nit: I don't think you need parens here?

pkg/kritis/secrets/secrets_test.go
ooq: Good point. Ack.

pkg/kritis/secrets/secrets_test.go
```diff
@@ -31,56 +31,99 @@ import (
 var pub, priv = createKeys("good")
 var pgpKey, err = NewPgpKey(priv, "", pub)
 
+var pub2, priv2 = passphraseProtectedSecretPublicKey, passphraseProtectedSecretPrivateKey
```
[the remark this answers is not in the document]
ooq: Basically this is the corresponding public key to the passphrase-protected private key. I'll put these data in a struct to reduce confusion.

PR #380, opened 2019-08-08
Final fixes to the standalone documentation for 0.1.0

aysylu:


aysylu:
> Something broke the kokoro build?

Yes, I filed https://github.com/zmap/zlint/issues/305 for the dependency that's broken, although during my investigation I couldn't find a quick way to fix it there.

Issue #387, opened 2019-08-12
Advertising use of key id instead of user id in docs


ooq:
**Is your feature request related to a problem? Please describe.**
We advertise the following workflow to generate key-pair using `gpg` (in docs, bug reproduction workflow, etc.):
```
gpg --quick-generate-key --yes my.attestor@example.com
gpg --armor --export my.attestor@example.com > gpg.pub
gpg --armor --export-secret-keys my.attestor@example.com > gpg.priv
```
This can cause reproduction issue because:

- `gpg` by default does substring match on provided email/user id. e.g.,  `attestor@example.com` is also a match for the above commands
- Multiple keys can be created for the same email/user id, thus a user might not exported the key intended.

**Describe the solution you'd like**
Update the doc to use key id instead of email/user id.

**Describe alternatives you've considered**
Or we can ask to first delete all keys under the user/email id that is to be used. A user might not want to delete.

**Additional context**
If possible, we should add instruction on key deletion in `clean up` section of the tutorials, to avoid key accumulation.

ooq:
A more reproducible steps could look like this:

1. ### Generate a new key with `gpg`.
```
gpg --quick-generate-key --yes my.attestor@example.com
```
The output should contain a 16-digit hexadecimal key id:
```console
$ gpg --quick-generate-key --yes my.attestor@example.com
...
gpg: key B8F04E5568F2C21B marked as ultimately trusted
...
```
2.  ### Save the hexadecimal key  to `KEYID`.
```
KEYID=B8F04E5568F2C21B
```
3.  ### Export the key using `KEYID`.
```
gpg --armor --export $KEYID > gpg.pub
gpg --armor --export-secret-keys $KEYID > gpg.priv
```

ooq:
Alternatively:

1. ### Generate a new key with `gpg`.
```
GPG_OUTPUT="$(gpg --quick-generate-key --yes attestor@example.com)"
```
2.  ### Extract the 40-digit fingerprint from output  to `KEY_FINGER_PRINT`.
```
echo $GPG_OUTPUT | sed -n 's/.*\([A-Z0-9]\{40\}\).*/\1/p'
```
3.  ### Export the key using `KEY_FINGER_PRINT`.
```
gpg --armor --export $KEY_FINGER_PRINT > gpg.pub
gpg --armor --export-secret-keys $KEY_FINGER_PRINT > gpg.priv
```

ooq:
Fixed by #388

PR #388, opened 2019-08-13
[ISSUE-387] Update doc and code with new way of exporting gpg key.

ooq:
Fixing #387.

### DIFF
--- a/docs/standalone/no_attestation.sh
+++ b/docs/standalone/no_attestation.sh
@@ -21,7 +21,7 @@ GPG_OUTPUT="$(gpg --quick-generate-key --yes attestor@example.com)"
 # Save its fingerprint.
 KEY_FINGER_PRINT="$(echo $GPG_OUTPUT | sed -n 's/.*\([A-Z0-9]\{40\}\).*/\1/p')"
 if [ ${#KEY_FINGER_PRINT} -ne 40 ]; then echo "Error: key fingerprnt is not 40 characters." ; exit
-else echo "Generated key finger print is $KEY_FINGER_PRINT."
+else echo "Generated key fingerprint is $KEY_FINGER_PRINT."
 fi
 
 gpg --armor --export $KEY_FINGER_PRINT gpg.pub
--- a/docs/standalone/no_attestation.sh
+++ b/docs/standalone/no_attestation.sh
@@ -24,7 +24,7 @@ if [ ${#KEY_FINGER_PRINT} -ne 40 ]; then echo "Error: key fingerprnt is not 40 c
 else echo "Generated key fingerprint is $KEY_FINGER_PRINT."
 fi
 
-gpg --armor --export $KEY_FINGER_PRINT gpg.pub
+gpg --armor --export $KEY_FINGER_PRINT > gpg.pub
 gpg --armor --export-secret-keys $KEY_FINGER_PRINT > gpg.priv
 
 PUBLIC_KEY=`base64 gpg.pub -w 0`
--- a/docs/standalone/no_attestation.sh
+++ b/docs/standalone/no_attestation.sh
@@ -19,13 +19,13 @@ set -e
 GPG_OUTPUT="$(gpg --quick-generate-key --yes attestor@example.com)"
 
 # Save its fingerprint.
-KEY_FINGER_PRINT="$(echo $GPG_OUTPUT | sed -n 's/.*\([A-Z0-9]\{40\}\).*/\1/p')"
-if [ ${#KEY_FINGER_PRINT} -ne 40 ]; then echo "Error: key fingerprnt is not 40 characters." ; exit
-else echo "Generated key fingerprint is $KEY_FINGER_PRINT."
+KEY_FINGERPRINT="$(echo $GPG_OUTPUT | sed -n 's/.*\([A-Z0-9]\{40\}\).*/\1/p')"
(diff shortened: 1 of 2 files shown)


codingandcoding:
[https://github.com/grafeas/kritis/tree/master/docs/standalone](url)

docs/standalone/no_attestation.sh
```diff
@@ -16,9 +16,16 @@
...
 
 # Create a public and private key pair.
-gpg --quick-generate-key --yes attestor@example.com
-gpg --armor --export attestor@example.com > gpg.pub
-gpg --armor --export-secret-keys attestor@example.com > gpg.priv
+GPG_OUTPUT="$(gpg --quick-generate-key --yes attestor@example.com)"
+
+# Save its fingerprint.
+KEY_FINGER_PRINT="$(echo $GPG_OUTPUT | sed -n 's/.*\([A-Z0-9]\{40\}\).*/\1/p')"
+if [ ${#KEY_FINGER_PRINT} -ne 40 ]; then echo "Error: key fingerprnt is not 40 characters." ; exit
```
aysylu: s/fingerprnt/fingerprint

aysylu: Why does the fingerprint need to be 40 chars? The error message as is would be confusing to users.

docs/standalone/no_attestation.sh
```diff
@@ -16,9 +16,16 @@
 set -e
 
 # Create a public and private key pair.
-gpg --quick-generate-key --yes attestor@example.com
-gpg --armor --export attestor@example.com > gpg.pub
-gpg --armor --export-secret-keys attestor@example.com > gpg.priv
+GPG_OUTPUT="$(gpg --quick-generate-key --yes attestor@example.com)"
+
+# Save its fingerprint.
+KEY_FINGER_PRINT="$(echo $GPG_OUTPUT | sed -n 's/.*\([A-Z0-9]\{40\}\).*/\1/p')"
```
aysylu: Could we use `gpg --fingerprint > fingerprint` instead here?

docs/standalone/no_attestation.sh
```diff
@@ -16,9 +16,16 @@
...
+if [ ${#KEY_FINGER_PRINT} -ne 40 ]; then echo "Error: key fingerprnt is not 40 characters." ; exit
+else echo "Generated key fingerprint is $KEY_FINGER_PRINT."
+fi
+
+gpg --armor --export $KEY_FINGER_PRINT > gpg.pub
```
aysylu: nit: let's keep this as one word, FINGERPRINT

docs/tutorial.md
```diff
@@ -34,10 +34,19 @@
     1. Create a public and private key pair. You'll be prompted to create a
        passphrase. Make sure to save it to the `PHRASE` environment variable.
 
+        Generate new key with `gpg` and extract its fingerprint.
         ```shell
-        gpg --quick-generate-key --yes attestor@example.com
-        gpg --armor --export attestor@example.com > gpg.pub
-        gpg --armor --export-secret-keys attestor@example.com > gpg.priv
+        GPG_OUTPUT="$(gpg --quick-generate-key --yes attestor@example.com)"
```
aysylu: Let's add a similar shell script that does all this to here? Otherwise, it's already not trivial to follow the tutorial.

docs/standalone/no_attestation.sh
ooq: ack

ooq: A GPG key fingerprint should be 40 chars. This is just to safeguard in case sed/gpg behaves differently across platform/version. Will improve the msg to be more general.

docs/standalone/no_attestation.sh
ooq: If we do something like `gpg --fingerprint attestor@example.com`. Then it will have the same problem as before.

docs/standalone/no_attestation.sh
ooq: ack

docs/tutorial.md
ooq: I think simplifying it further is definitely a good idea. I think it will probably worth a separate PR? (we should also consider reusing user's existing `attestor@example.com` key, add step to delete key, etc). Leaving these commands in the tutorial for now probably gives user more transparency when the clean-up scripts, etc are not in place yet?

aysylu: I'm fine with adding this in a separate PR. I'd rather not assume that a user who ran 1 of the tutorials will run the other, and in which order. We can just change the usernames here, if it helps.

ooq: We agree that we should not assume the order that the tutorials run. With using of fingerprint username is probably not an issue.

docs/standalone/no_attestation.sh
aysylu: I think this is the only one left, maybe you have an unexported commit?

ooq: hmm, It should have been fixed in this [commit](https://github.com/grafeas/kritis/pull/388/commits/983356e6b8ed20c5a40ac7f68c80418996b64a3e). Could you check the overall file changes? I should probably have hit `resolve` though.

aysylu:
I see it now, strange! Thanks for resolving.:)

Issue #389, opened 2019-08-13
Base64 encoding in no_attestation.sh fails for MacOX


ooq:
## Expected Behavior
no_attestation.sh runs under MacOX

## Actual Behavior
`base64: invalid option -- w`

## Steps to Reproduce the Problem

MacOX
`./no_attestation.sh`

## Additional info
Add any other context about the problem here (e.g. screenshots, links)

ooq:
Fixed by #390 .

PR #390, opened 2019-08-13
[ISSUE-389] Fix base64 encoding error for Mac OX.

ooq:
Fixing #389


docs/standalone/no_attestation.sh
```diff
@@ -20,7 +20,11 @@ gpg --quick-generate-key --yes attestor@example.com
 gpg --armor --export attestor@example.com > gpg.pub
 gpg --armor --export-secret-keys attestor@example.com > gpg.priv
 
-PUBLIC_KEY=`base64 gpg.pub -w 0`
+if [ "$(uname)" == "Darwin" ]; then
```
aysylu: nit: please add a short comment that this is for MacOs

Issue #391, opened 2019-08-21
Kritis leaks memory


klauern:
## Expected Behavior
It runs at a stable memory footprint

## Actual Behavior
it doesn't?

## Steps to Reproduce the Problem
Run it for a while

## Additional info

I don't know how to better debug the issue, but here's a graph:
![image](https://user-images.githubusercontent.com/4735/63433240-6c668d80-c3e8-11e9-8b06-49ac32f0e9a1.png)

Each of these is an individual container on a separate Kubernetes cluster, so they are similar in nature but will have their own footprint.

I'd appreciate any guidance on what I can do to better provide information on what is going on.  This is my only window into the view that I can find, and I'm scratching my head to find other ways to get this information on a running cluster.

klauern:
Note that the one that looks the most stable is our least-used server, so it doesn't get any traffic, hence it doesn't really fluctuate.

ooq:
Hi @klauern , can you describe the workload a little bit?

vbanthia:
We also faced the same issue and I did some profiling and found out memory leak is because of grafeasclient connections are not closed once the request is handled properly. 

I created debug docker image with netstat installed and observe new grpc connections (TCP) getting created whenever a new pod is created. These connections do not get closed and left open. 

```console
root@kritis-validation-hook-f76fd4c75-z2cvg:/go# netstat -W | wc -l
524
```

**Connections**
```
tcp        0      0 kritis-validation-hook-f76fd4c75-z2cvg:55436 nrt12s23-in-f10.1e100.net:https ESTABLISHED
tcp        0      0 kritis-validation-hook-f76fd4c75-z2cvg:44904 nrt12s13-in-f10.1e100.net:https ESTABLISHED
tcp        0      0 kritis-validation-hook-f76fd4c75-z2cvg:54140 nrt12s15-in-f74.1e100.net:https ESTABLISHED
```

We internally use kritis fork and temporary using this fix. 
https://github.com/mercari/kritis/pull/18


Ideally, Kritis should not create grafeasclient inside handler but instead should be initialized before calling request handler so that same connection can be reused between goroutines (request handlers)

(https://github.com/grafeas/kritis/blob/681e6d3e8a4675d386e580fd643304c5e0263245/pkg/kritis/admission/admission.go#L246)

aysylu:
[...]

PR #392, opened 2019-08-21
Add pprof profiling to allow us to debug memory issues.

clrprod:
This should allow us to investigate bugs like #391 by requesting a dump of memory allocations


aysylu:
Help me understand, would this affect Kritis running in prod? If so, is there any way to conditionally enable this?

clrprod:
It shouldn't have an impact - it adds some debugging endpoints that gather information if you hit them. This code runs in prod all over the place.

PR #393, opened 2019-08-22
Fix memory leak

vbanthia:
## What

Fixes memory leak issue. 

In current Kritis implementation, GrafeasClient is created inside the request handler and they are not closed gracefully after the request has been handled. This creates memory leak. 

A better approach might be to create grafeasclient before handling requests and use same client for all goroutines (request handler) but this will require more refactoring. This is a quick fix and works well.


vbanthia:
Wrong upstream. Sorry.

aysylu:
I'd be happy to accept a PR with the fix to this project, @vbanthia

PR #394, opened 2019-08-22
[Issue-391] - Fixes memory leak issue by closing unused connections

vbanthia:
Fixes Issue https://github.com/grafeas/kritis/issues/391

NOTE: This is not a clean solution. Ideally, Kritis should not create upstream client inside handler. It should be created outside and passed as an argument to request handlers. This way same client can be used by all goroutines. This will require a bit of refactoring, till then we can use this solution to fix memory issues.

### DIFF
--- a/pkg/kritis/metadata/containeranalysis/cache.go
+++ b/pkg/kritis/metadata/containeranalysis/cache.go
@@ -45,6 +45,11 @@ func NewCache() (*Cache, error) {
 	}, nil
 }
 
+// Close closes client connections
+func (c Cache) Close() {
+	c.client.Close()
+}
+
 // Vulnerabilities gets Package Vulnerabilities Occurrences for a specified image.
 func (c Cache) Vulnerabilities(image string) ([]metadata.Vulnerability, error) {
 	if v, ok := c.vuln[image]; ok {
--- a/pkg/kritis/metadata/containeranalysis/containeranalysis.go
+++ b/pkg/kritis/metadata/containeranalysis/containeranalysis.go
@@ -58,6 +58,11 @@ func New() (*Client, error) {
 	}, nil
 }
 
+// Close closes client connections
+func (c Client) Close() {
+	c.client.Close()
+}
+
 //Vulnerabilities gets Package Vulnerabilities Occurrences for a specified image.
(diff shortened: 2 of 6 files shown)


pkg/kritis/metadata/containeranalysis/cache.go
```diff
@@ -45,6 +45,11 @@ func NewCache() (*Cache, error) {
 	}, nil
 }
 
+// Close closes client connections
```
ooq: nit: typo

pkg/kritis/metadata/containeranalysis/containeranalysis.go
```diff
@@ -58,6 +58,11 @@ func New() (*Client, error) {
 	}, nil
 }
 
+// Close closes client connections
```
ooq: nit: typo

pkg/kritis/metadata/grafeas/grafeas.go
```diff
@@ -111,6 +111,12 @@ func New(config kritisv1beta1.GrafeasConfigSpec, certs *CertConfig) (*Client, er
 	}, nil
 }
 
+// Close closes client connections
```
ooq: nit: typo

pkg/kritis/metadata/metadata.go
```diff
@@ -35,6 +35,8 @@ type Fetcher interface {
 	CreateAttestationNote(aa *kritisv1beta1.AttestationAuthority) (*grafeasv1beta1.Note, error)
 	//Attestations get Attestation Occurrences for given image.
 	Attestations(containerImage string) ([]PGPAttestation, error)
+	// Close closes client connections
```
ooq: nit: typo

pkg/kritis/metadata/containeranalysis/cache.go
vbanthia: Hi @ooq Thanks for the review. I can't see any typo here. Is there any grammar mistake in this comment. Please let me know.

vbanthia:
> Is it possible to add some tests?

I thought about this and still can't come up with any better idea to test zombie connections with unit tests. Mostly, memory-related scenarios get tested in benchmarking.

Please let me know if you have any easy idea to test it.

PR #398, opened 2019-09-06
Use namespaceSelector in kritis webhook to whitelist important namespaces such as kube-system

vbanthia:
## WHAT
This will allow user to whitelist some namespace where they do not want to run Kritis. By this, admission controller will not send validation request where namespace has label `kritis-validation: disabled`. 

**NOTE:** This feature is different from break-glass. In case of break-glass, request still goes to kritis and kritis ignore the validation where as with this admission controller itself will never send request to kritis webhook server. 

## WHY

Currently kritis webhook is enabled for all namespaces which can create deadlock situation. With this feature, we can whitelist some namespace such as kube-system and namespace where kritis is running so that the deadlock situation can be prevented. 

Without this kritis need to be always available else it can create a deadlock where no new pod can be created including kritis itself.

### DIFF
--- a/helm-hooks/postinstall/postinstall.go
+++ b/helm-hooks/postinstall/postinstall.go
@@ -85,6 +85,9 @@ webhooks:
         resources:
           - pods
     failurePolicy: Fail
+    namespaceSelector:
+      matchExpressions:
+      - {key: kritis-validation, operator: NotIn, values: [disabled]}
     clientConfig:
       caBundle: %s
       service:
@@ -119,6 +122,9 @@ webhooks:
           - deployments
           - replicasets
     failurePolicy: Fail
+    namespaceSelector:
+      matchExpressions:
+      - {key: kritis-validation, operator: NotIn, values: [disabled]}
     clientConfig:
       caBundle: %s
       service:


vbanthia:
@aysylu 

Sorry for the late response. 

> To make sure I understand the issue this PR is addressing, could you please give an example of a scenario in which global image allowlist is not sufficient? 


Below is pod creation request flow:

```
User --create-pod--> kube-api (admission controller) --validate-pod--> Kritis
```

In case Kritis pod is not reachable because of some issue such as node where kritis pod is running is not reachable. So, when admission controller will try to make validation request, it will always fail. No matter what you have configured in `global image allowlist`.

Error example:

```
FailedCreate  16s (x69 over 10h)  daemonset-controller  Error creating: Internal error occurred: failed calling webhook "kritis-validation-hook.grafeas.io": Post https://kritis-validation-hook.kritis-dev.svc:443/?timeout=30s: no endpoints available for service "kritis-validation-hook
```

Once you have this situation, you cannot even start a new pod. It is like a deadlock. Only workaround after this point is to delete kritis validationwebhook itself. 


There are two solutions for this problem. 

#### 1. Use `failurePolicy=Ignore` for kritis ValidatingWebhookConfiguration. 

https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy

This will ignore webhook validation in case of timeout or other unrecognized errors.

#### 2. Use namespace selector for kritisValidationWebhook

Using namespace selector, we can label some important namespaces such as `kube-system` or `kritis` to ignore validation requests for these namespaces. Request flow for these namespaces will be:

```
User --create-pod--> kube-api (admission controller)
```

Request will not go to Kritis webserver and so kritis availability does not matter for these critical namespaces. So, there will not be any deadlock situation. 

Istio also uses namespace selector for their mutatingwebhook. 

https://github.com/istio/istio/blob/master/install/kubernetes/helm/istio/charts/sidecarInjectorWebhook/templates/mutatingwebhook.yaml#L24-L34

aysylu:
@vbanthia, thank you for your thorough response and the excellent examples. This PR looks great. Would you be willing to contribute documentation on this in the docs folder, for future reference? I'd be happy to review a follow-up PR on this. Thanks!

vbanthia:
@aysylu 

Thanks. I will add documentation soon for this in this PR.

aysylu:
@vbanthia sounds great, thanks! I'm happy to merge this as is and have the documentation in a follow-up PR. Merging this one.

PR #407, opened 2019-10-08
Change AAList to single AA in ISP.

ooq:
Attestation authority in ISP is used for signing image that passes vuln criteria. There is no need to have multiple AAs in ISP.
This change is also a prerequisite for moving secret key from AA to ISP.

PR #408, opened 2019-10-09
Change AAList to single AA in ISP.

ooq:
Attestation authority in ISP is used for signing image that passes vuln criteria. There is no need to have multiple AAs in ISP.
This change is also a prerequisite for moving secret key from AA to ISP.

This PR is part of bundled change to achieve better compatibility with Binary Authorization.

### DIFF
--- a/artifacts/examples/generic-attestation-policy-example.yaml
+++ b/artifacts/examples/generic-attestation-policy-example.yaml
@@ -4,5 +4,4 @@ metadata:
   name: my-gap
   namespace: default
 spec:
-  attestationAuthorityNames:
-  - kritis-authority
+  attestationAuthorityName: kritis-authority
--- a/artifacts/examples/generic-attestation-policy-example.yaml
+++ b/artifacts/examples/generic-attestation-policy-example.yaml
@@ -4,4 +4,5 @@ metadata:
   name: my-gap
   namespace: default
 spec:
-  attestationAuthorityName: kritis-authority
+  attestationAuthorityNames:
+  - kritis-authority
--- a/pkg/kritis/review/review.go
+++ b/pkg/kritis/review/review.go
@@ -104,9 +104,13 @@ func (r Reviewer) ReviewISP(images []string, isps []v1beta1.ImageSecurityPolicy,
 	for _, isp := range isps {
 		glog.Infof("Validating against ImageSecurityPolicy %s", isp.Name)
 		// Get all AttestationAuthorities in this policy.
-		auths, err := r.getAttestationAuthoritiesForISP(isp)
+		a, err := r.getAttestationAuthorityForISP(isp)
 		if err != nil {
 			return err
+		}
+		auths := make([]v1beta1.AttestationAuthority, 0)
+		if a != nil {
+			auths = append(auths, *a)
 		}
 		for _, image := range images {
 			glog.Infof("Check if %s as valid Attestations.", image)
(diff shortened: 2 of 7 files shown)


ooq:
@aysylu kokoro build somehow links to #407, do you know why?

ooq:
@aysylu never mind, it's probbaly some strange caching behavior. It's gone after I made some changes.

pkg/kritis/review/review.go
```diff
@@ -103,15 +103,18 @@ func (r Reviewer) ReviewISP(images []string, isps []v1beta1.ImageSecurityPolicy,
...
 		if err != nil {
 			return err
 		}
+
 		for _, image := range images {
 			glog.Infof("Check if %s as valid Attestations.", image)
-			notAttestedBy := r.findUnsatisfiedAuths(image, auths)
-			imgAttested := len(notAttestedBy) == 0
+			imgAttested := false
+			if auth != nil {
```
nenaddedic: It should never be nil, right (based on L107-L110).  So perhaps this should return an error if auth == nil.

pkg/kritis/review/review.go
```diff
@@ -174,42 +185,40 @@ func (r Reviewer) handleViolations(image string, pod *v1.Pod, violations []polic
 	return fmt.Errorf(errMsg)
 }
 
-// Create attestations for 'image' by all 'auths'.
-func (r Reviewer) addAttestations(image string, isp v1beta1.ImageSecurityPolicy, auths []v1beta1.AttestationAuthority) error {
+// Create attestation for 'image' by ISP auth.
+func (r Reviewer) addAttestations(image string, isp v1beta1.ImageSecurityPolicy, auth v1beta1.AttestationAuthority) error {
```
nenaddedic: Change name to singular:
addAttestation()

pkg/kritis/review/review.go
ooq: There is a special case where AA name is empty string "". In this case, should we throw error much earlier on?

nenaddedic: You mean when isp.Name is empty?

ooq: I mean isp.AttestationAuthorityName

nenaddedic: Oh, I think I understand now what is going on.  If you do not specify an AA in isp, then the isp decision is not "cached" as an attestation, and it is instead re-evaluated every time.  This sounds intentional, but I'm not sure it is documented anywhere.  We should just leave it as is, but document better.  WDYT?

artifacts/examples/generic-attestation-policy-example.yaml
```diff
@@ -4,5 +4,4 @@ metadata:
   name: my-gap
   namespace: default
 spec:
-  attestationAuthorityNames:
-  - kritis-authority
+  attestationAuthorityName: kritis-authority
```
steakunderscore: This file is an example of a GAP. From reading the comments in the PR, this should only be changing ISPs. Perhaps this is a mistake?

ooq: Yes, thanks for pointing out! Will fix.

pkg/kritis/review/review.go
ooq: Sure. I'll update this PR with doc.

PR #409, opened 2019-10-11
Add allowlist support to Kritis GAP

ooq:
Allowlist admission name pattern follows: https://cloud.google.com/binary-authorization/docs/policy-yaml-reference#admissionwhitelistpatterns


ooq:
@nenaddedic PTAL, thanks.

pkg/kritis/apis/kritis/v1beta1/genericattestation.go
```diff
@@ -31,9 +31,14 @@ type GenericAttestationPolicy struct {
 	Spec GenericAttestationPolicySpec `json:"spec"`
 }
 
+type AdmissionWhitelistPatternSpec struct {
```
nenaddedic: I think it's "allowlist" in Kritis codebase.

pkg/kritis/review/review.go
```diff
@@ -52,12 +52,26 @@ func New(client metadata.Fetcher, c *Config) Reviewer {
...
+	}
+	return allowlist
+}
+
 // ReviewGAP reviews images against generic attestation policies
 // Returns error if violations are found and handles them per violation strategy
 func (r Reviewer) ReviewGAP(images []string, gaps []v1beta1.GenericAttestationPolicy, pod *v1.Pod) error {
 	images = util.RemoveGloballyAllowedImages(images)
+
+	images = util.RemoveGapAllowedImages(images, generateGapAllowlist(gaps))
```
nenaddedic: It is odd to flatten allowlists from different policies into a single allowlist.  Allowlist should be handled separately for each policy.  On the other hand the outer loop below is by image, so not clear what to do.

We could invert the loops below (by gap first, by image then), and then allowlist handling can be per policy.  WDYT?

pkg/kritis/review/review.go
```diff
@@ -52,12 +52,26 @@ func New(client metadata.Fetcher, c *Config) Reviewer {
 	}
 }
 
+// Flatten allowlist from a list of generic attestation policies
+func generateGapAllowlist(gaps []v1beta1.GenericAttestationPolicy) []string {
```
nenaddedic: This should really take a single GAP, as it doesn't make much sense to have a single flattened allowlist from multiple policies.  On the other hand, I understand that there are some problems with this, please see my other comment below.

pkg/kritis/util/allowlist.go
```diff
@@ -39,20 +41,93 @@ func RemoveGloballyAllowedImages(images []string) []string {
...
+			glog.Errorf("couldn't check if %s is in gap allowlist: %v", image, err)
+		}
+		if !allowlisted {
+			notAllowlisted = append(notAllowlisted, image)
+		}
+	}
+	return notAllowlisted
+}
+
+// Do an image match based on reference.
```
nenaddedic: What's a reference?

pkg/kritis/util/allowlist.go
```diff
@@ -39,20 +41,93 @@ func RemoveGloballyAllowedImages(images []string) []string {
...
+
+// Do an image match based on name pattern.
+// See https://cloud.google.com/binary-authorization/docs/policy-yaml-reference#admissionwhitelistpatterns
+func imageNamePatternMatch(image string, pattern string) (bool, error) {
+	if len(pattern) == 0 {
+		return false, errors.New("empty pattern")
+	}
+	if pattern[len(pattern)-1] == '*' {
+		pattern = pattern[:len(pattern)-1]
+		if strings.HasPrefix(image, pattern) {
```
nenaddedic: Optional: you can combine the two ifs into a single 'if a && b'.

pkg/kritis/util/allowlist.go
```diff
@@ -39,20 +41,93 @@ func RemoveGloballyAllowedImages(images []string) []string {
...
+		return false, err
+	}
+	// Make sure images have the same context
+	if reflect.DeepEqual(allowRef.Context(), imageRef.Context()) {
+		return true, nil
+	}
+	return false, nil
+}
+
+// Do an image match based on name pattern.
```
nenaddedic: Could you document what the matching rules are?

pkg/kritis/apis/kritis/v1beta1/genericattestation.go
ooq: so we are not adopting the name from BinAuthz?

pkg/kritis/review/review.go
ooq: The biggest reason in doing it this way now is now to have kritis api dependency in `util` -- which I don't have strong opinion on. Or we can have this function in `genericattestation.go`.

pkg/kritis/apis/kritis/v1beta1/genericattestation.go
nenaddedic: I would guess not, but let's confirm with Aysylu.

pkg/kritis/review/review.go
nenaddedic: I did not follow this.  Which dependency are you trying to avoid?

pkg/kritis/apis/kritis/v1beta1/genericattestation.go
ooq: Let‘s stick with allowlist for now then.

pkg/kritis/util/allowlist.go
ooq: basically the way allowlist is handled in kritis is based on `name.ParseReference` and compare the references. For this newly added allowlist in GAP, i'm going to follow BinAuthz and use prefix matching. But maintain this reference-based matching for ISP.

pkg/kritis/review/review.go
nenaddedic: Let's postpone this until ReviewGAP() is restructured as discussed (for loops inverted).  LGTM as is.

ooq: Resolving this now since the loop inversion is only possible after #432 is addressed.

PR #410, opened 2019-10-11
Fix a logic bug in the GenericAttestationPolicy review.

nenaddedic:
Currently intended semantics is disjunctive: an image passes policy
checks if at least one GAP is satisfied, which in turn is if at least
one AttestationAuthority verifies the signature.

### DIFF
--- a/pkg/kritis/review/review.go
+++ b/pkg/kritis/review/review.go
@@ -68,7 +68,7 @@ func (r Reviewer) ReviewGAP(images []string, gaps []v1beta1.GenericAttestationPo
 
 	for _, image := range images {
 		glog.Infof("Check if %s has valid Attestations.", image)
-		var imgAttested bool
+		imgAttested := false
 		for _, gap := range gaps {
 			glog.Infof("Validating against GenericAttestationPolicy %s", gap.Name)
 			// Get all AttestationAuthorities in this policy.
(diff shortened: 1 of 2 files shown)


aysylu:
Kokoro test failure is due to an intermittent timeout, unrelated to this PR. I'll re-run it once the branch is synced with HEAD and the minor comments are addressed.

steakunderscore:
What's the design decision for this? Reading the ticket I assume this was discussed internally, perhaps exposing some of that communication would be useful.

We currently make use of Binauth's `requireAttestationsBy` where we require 3 attestors to be satisfied before we allow images to be admitted. With this change, is there a way that an admission can require multiple attestors?

For context, from the binauth [policy ref]:
> `requireAttestationsBy` specifies one or more attestors who must authorize the release before a container image can be deployed.

Let me know if this would be better discussed somewhere else. I can't find the Kritis Google Group, only the Google Groups for Grafeas.

[policy ref]: https://cloud.google.com/binary-authorization/docs/policy-yaml-reference

nenaddedic:
Henry, yes, this ended up more confusing than necessary, sorry about that!

First, some reassurance: As indicated in this post
<https://groups.google.com/forum/#!topic/kritis-users/ydZkjo39n-c> (btw
that's the Kritis Google Group), our plan is to implement the semantics
which you want -- ALL attestors required.

Second, Kritis does not currently implement the Binary Auth semantics,
instead it wants ANY attestor.  I had introduced a bug where I accidentally
changed that to ALL attestors.

Third, this PR restores the previous ANY attestor semantics, with full
awareness that this is changing soon.  The change to ALL attestors is going
to be part of a "package" of changes (see referenced post), so we don't
want to sneak it in and break backwards compatibility unannounced.

We realize that the discrepancy between Binary Auth semantics and Kritis
semantics is confusing, and are actively working on fixing that.  Thanks
for your understanding and, as always, PRs are welcome :)

aysylu:
@nenaddedic: +1 to what you said, thanks for the clarifications!

@steakunderscore: I just realized the group's visiblity settings made it available only to the group's members. Now fixed: https://groups.google.com/forum/#!forum/kritis-users. Please feel free to join!

aysylu:
I'm addressing the kokoro test failure and looking into it separately, as it's not caused by this PR.

PR #399, opened 2019-09-06
[Issue #396] Add support for the separation of image and Attestor into different GCP projects

marco-lancini:
Add support for the separation of GCR and Attestor into different GCP projects

PR #413, opened 2019-10-16
v0.2.0 release

aysylu:
See CHANGELOG for full details.


aysylu:
I found the culprit for the failing kokoro test, in `containeranalysis_int_test`, introduced by #399. Fix is on the way, then we can release this after syncing to HEAD.

aysylu:
The failed test is flaky and passes locally.

Issue #396, opened 2019-09-04
No support for the separation of GCR and Attestor into different GCP projects


marco-lancini:
## Expected Behavior

It would be expected that Kritis would respect the value of `noteReference` spec of a `AttestationAuthority` both when creating and fetching attestations.

## Actual Behavior

Although when creating attestations it respects the `noteReference` value [1], when listing attestations kritis uses the project in which the GCR lives, by extracting it directly from the image name [2].
This prevent the case of having GCR and the Attestor living in 2 different GCP projects.

[1] https://github.com/grafeas/kritis/blob/bc811c62913cdc0b95ce59949084dad42d6c9e4e/pkg/kritis/metadata/containeranalysis/containeranalysis.go#L149
[2] https://github.com/grafeas/kritis/blob/bc811c62913cdc0b95ce59949084dad42d6c9e4e/pkg/kritis/metadata/containeranalysis/containeranalysis.go#L102 

## Steps to Reproduce the Problem

1. Install and run Kritis as explained in the [installation guide](https://github.com/grafeas/kritis/blob/master/docs/install.md)
2. The projects used are:
    A. `project-gcr`: hosting GCR
    B. `project-attestor`: hosting the attestor
3. Create an attestation authority (note the `noteReference: v1beta1/projects/project-attestor`)
```
---
apiVersion: v1
kind: List
items:
  - apiVersion: kritis.grafeas.io/v1beta1
    kind: GenericAttestationPolicy
    metadata:
      name: my-gap
      namespace: default
      labels:
        kritis.grafeas.io/install: ''
    spec:
      attestationAuthorityNames:
        - kritis-authority
  - apiVersion: kritis.grafeas.io/v1beta1
    kind: AttestationAuthority
    metadata:
      name: kritis-authority
      namespace: default
    spec:
      noteReference: v1beta1/projects/project-attestor
      privateKeySecretName: gcp-attestations-pgp
      publicKeyData: /secret/gpg.pub
      volumeMounts:
        - name: gcp-attestations-pgp
          mountPath: /secret
      volumes:
        - name: gcp-attestations-pgp
          secret:
            secretName: gcp-attestations-pgp
```

4. Try to deploy an image which has attestations and should be allowed (note `project-gcr` used in `image`):
```
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Pod
metadata:
  name: test-alpine
spec:
  containers:
    - name: test-alpine
[...]

PR #414, opened 2019-10-16
Fixed integration to use NoteReference in AttestationAuthority for attaching occurrences

aysylu:
This addresses integration issues that were uncovered when we moved to using `ListNoteOccurrences` in #399, and completes the implementation for #396.

* Removed API version from AttestationAuthority's `NoteReference`.
* Fixed note name in the `ListNoteOccurrencesRequest`.
* Fixed retrieval of note occurrences logic.
* s/Occurence/Occurrence
* moved `GetProjectFromNoteReference` to `metadata`, which is lowest common ancestor, to avoid code duplication

### DIFF
--- a/integration/crd.go
+++ b/integration/crd.go
@@ -37,7 +37,7 @@ kind: AttestationAuthority
 metadata:
   name: test-attestor
 spec:
-  noteReference: projects/krits-int-test
+  noteReference: projects/kritis-int-test
   privateKeySecretName: %s
   publicKeyData: %s`
 )
--- a/integration/crd.go
+++ b/integration/crd.go
@@ -37,7 +37,7 @@ kind: AttestationAuthority
 metadata:
   name: test-attestor
 spec:
-  noteReference: v1alpha1/projects/krits-int-test
+  noteReference: projects/krits-int-test
   privateKeySecretName: %s
   publicKeyData: %s`
 )
--- a/pkg/kritis/metadata/containeranalysis/containeranalysis.go
+++ b/pkg/kritis/metadata/containeranalysis/containeranalysis.go
@@ -19,7 +19,6 @@ package containeranalysis
 import (
 	"fmt"
 	"strings"
-	"time"
 
 	ca "cloud.google.com/go/containeranalysis/apiv1beta1"
 	"github.com/golang/glog"
@@ -81,6 +80,10 @@ func (c Client) Vulnerabilities(containerImage string) ([]metadata.Vulnerability
 }
 
 //Attestations gets AttesationAuthority Occurrences for a specified image, using the note specified in the AttestationAuthority provided.
+// This may take a few seconds to retrieve an attestation occurrence, if it was created very recently.
+// For GenericAttestationPolicy, this has little impact as it's expected that attestations will be created before a pod admission request is sent.
+// For ImageSecurityPolicy, which effectively caches the previous policy decision in an attestation, the policy will be re-evaluated if an attestation occurrence has not yet been retrieved.
+// In most cases, it's expected that ImageSecurityPolicy will return the same decision, as vulnerability scannig process takes longer than a few seconds to run and update metadata.
 func (c Client) Attestations(containerImage string, aa *kritisv1beta1.AttestationAuthority) ([]metadata.PGPAttestation, error) {
 	var p []metadata.PGPAttestation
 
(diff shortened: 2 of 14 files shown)


aysylu:
@marco-lancini this addresses the problems that were discovered in integration tests (plus, some other long-overdue minor fixes to the codebase), and unblocks the release of `v0.2.0` with the feature you contributed! Once the PR is reviewed and merged, I'll release Kritis `v0.2.0` right away. Thanks again for your patience!

marco-lancini:
Sounds great, thanks! 😃

pkg/kritis/metadata/containeranalysis/containeranalysis.go
```diff
@@ -128,29 +129,53 @@ func (c Client) fetchAttestationOccurrence(containerImage string, kind string, a
 		return nil, fmt.Errorf("%s is not a valid image hosted in GCR", containerImage)
 	}
 
+	noteName := fmt.Sprintf("%s/notes/%s", auth.Spec.NoteReference, auth.Name)
```
nenaddedic: NoteReference should IMO contain the entire note name, like in https://cloud.google.com/binary-authorization/docs/reference/rest/v1/projects.attestors.  The name of the note should not be derived from auth.Name.

For example, an AttestationAuthority might look like:
Name: 'authority-qa'
Spec:
  NoteReference: 'projects/some-secure-project/notes/qa-note-123'

pkg/kritis/metadata/containeranalysis/containeranalysis.go
```diff
@@ -128,29 +129,53 @@ func (c Client) fetchAttestationOccurrence(containerImage string, kind string, a
...
 	req := &grafeas.ListNoteOccurrencesRequest{
-		Name: auth.Spec.NoteReference,
-		// TODO: re-add `kind` clause in filter
-		//       (currently not supported as per https://github.com/grafeas/kritis/issues/401#issuecomment-538947608)
+		Name: noteName,
```
nenaddedic: As per comment above, I suggest reverting the change to this line.

pkg/kritis/metadata/containeranalysis/containeranalysis.go
```diff
@@ -128,29 +129,53 @@ func (c Client) fetchAttestationOccurrence(containerImage string, kind string, a
...
 		// 		Filter:  fmt.Sprintf("resourceUrl=%q AND kind=%q", util.GetResourceURL(containerImage), kind),
 		Filter:   fmt.Sprintf("resourceUrl=%q", util.GetResourceURL(containerImage)),
 		PageSize: constants.PageSize,
 	}
 
-	it := c.client.ListNoteOccurrences(c.ctx, req)
+	return c.listNoteOccurrences(req)
+}
+
+func (c Client) listNoteOccurrences(req *grafeas.ListNoteOccurrencesRequest) ([]*grafeas.Occurrence, error) {
```
nenaddedic: I don't really understand how this function works:
1. Where is pagination handled?
2. What is the purpose of fetching Occurrences only when a clock tick happens?

pkg/kritis/metadata/containeranalysis/containeranalysis.go
aysylu: I agree. Could you please file an issue for this? This definitely fits with the other work you and @ooq are doing for compatibility.

pkg/kritis/metadata/containeranalysis/containeranalysis.go
aysylu: Let's do this as part of the larger change to have consistency with BinAuthz on how attestors represent the note.

pkg/kritis/metadata/containeranalysis/containeranalysis.go
aysylu: Because listing note occurrences gives results with a staleness bound, the goal of this function is to give the opportunity for the newly created attestations to be retrieved from `containeranalysis`. To answer your specific questions:

1. The client doesn't expose pagination logic, I suspect because it's handled internally already: [reference](https://godoc.org/cloud.google.com/go/containeranalysis/apiv1beta1#GrafeasV1Beta1Client.ListNoteOccurrences). I filed #419 to verify this is actually the case.

2. This is to ensure we don't send too many `ListNoteOccurrences` requests in a given time unit. Because of the staleness bound, there's no point to actively be checking if occurrences have propagated through the system after receiving none.

Question for you @nenaddedic: in BinAuthz, do you expect to have multiple occurrences per attestor?

nenaddedic: Multiple occurrences per image per attestor would be uncommon, but possible.  (I assume you meant "per image per attestor".)

Polling for attestations has its advantages, but I think they're outweighed by disadvantages:
1. The common "no attestations" rejection now necessarily takes 20 seconds to reach.
2. Code clarity.  It is not clear why keep on polling only when "no occurrences" is returned.  Timeouts (L154) should arguably be handled in this way as well.  Maybe also unverifiable attestations.

I suggest removing the polling here and changing the failing integration test to retry for 20 seconds.  (Unless that turns out to be more work than you can afford now.)

aysylu: We discussed this in more detail yesterday: @nenaddedic made good points, and we agreed on the approach. I added the comments to reflect our current understanding.

integration/crd.go
```diff
@@ -37,7 +37,7 @@ kind: AttestationAuthority
 metadata:
   name: test-attestor
 spec:
-  noteReference: v1alpha1/projects/krits-int-test
+  noteReference: projects/krits-int-test
```
nenaddedic: "krits-..." or "kritis-..." ?
(letter "i")

pkg/kritis/metadata/containeranalysis/containeranalysis.go
```diff
@@ -80,6 +80,10 @@ func (c Client) Vulnerabilities(containerImage string) ([]metadata.Vulnerability
 }
 
 //Attestations gets AttesationAuthority Occurrences for a specified image, using the note specified in the AttestationAuthority provided.
+// This may take a few seconds to retrieve an attestation occurrence, if it was created very recently.
```
nenaddedic: Thanks for the clarifying comment!

pkg/kritis/metadata/containeranalysis/containeranalysis.go
nenaddedic: Filed https://github.com/grafeas/kritis/issues/420

pkg/kritis/metadata/containeranalysis/containeranalysis.go
aysylu: 👍

integration/crd.go
aysylu: good catch, fixed

Issue #420, opened 2019-11-07
AttestationAuthority.Spec.NoteReference should contain the entire note name


nenaddedic:
**Is your feature request related to a problem? Please describe.**
Current implementation is inconsistent with Binary Authorization, for no good reason.  The Note that an AttestationAuthority uses is derived from two things: (i) auth.Name, (ii) auth.Spec.noteReference (which is effectively just the project name).  This makes it impossible to fully decouple the Note from the AttestationAuthority.

**Describe the solution you'd like**
auth.Spec.noteReference should be required to contain the entire note name like "projects/foo/notes/bar".

**Additional context**
https://github.com/grafeas/kritis/pull/414

Issue #326, opened 2019-02-12
Kokoro test failures unrelated to the PR


aysylu:
## Expected Behavior
Tests pass when the functionality is correct.

## Actual Behavior
Potentially flaky tests. Failures in #324 and #325 that are unrelated to the PRs. The integration tests pass locally.

## Steps to Reproduce the Problem
Create a PR with non-code changes and see if kokoro tests pass.

aysylu:
Some investigation was done in #327, but unable to successfully reproduce due to most likely merging the fix in #324. Will re-open the investigation if subsequent PRs are having this issue.

aysylu:
This is happening again, with 2 different types of failures in #329.

PR #423, opened 2019-11-08
Define a read-only metadata client and use it in GAP

nenaddedic:
This splits the general-purpose metadata.Fetcher into two:

- metadata.ReadOnlyClient: used in GAP, cannot mutate attestation storage, narrow interface
- metadata.ReadWriteClient: used in ISP and signer, has methods to mutate attestation storage

This PR will also make it easier to address https://github.com/grafeas/kritis/issues/420 by making it clear where and how metadata.AttestationNote() is used (notably: not in GAP review).

### DIFF
--- a/pkg/kritis/admission/admission_test.go
+++ b/pkg/kritis/admission/admission_test.go
@@ -115,13 +115,16 @@ func Test_AdmissionResponse(t *testing.T) {
 	}
 	for _, tc := range tcs {
 		t.Run(tc.name, func(t *testing.T) {
-			mReviewer := func(client metadata.Fetcher) reviewer {
+			mReviewer := func() reviewer {
 				return testutil.NewReviewer(tc.reviewGAPErr, tc.reviewISPErr, tc.expectedMsg)
 			}
 			mockConfig := config{
 				retrievePod: mockValidPod(),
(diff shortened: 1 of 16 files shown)


pkg/kritis/admission/admission_test.go
```diff
@@ -115,14 +115,17 @@ func Test_AdmissionResponse(t *testing.T) {
...
 		t.Run(tc.name, func(t *testing.T) {
-			mReviewer := func(client metadata.Fetcher) reviewer {
+			mReviewer := func() reviewer {
 				return testutil.NewReviewer(tc.reviewGAPErr, tc.reviewISPErr, tc.expectedMsg)
 			}
 			mockConfig := config{
 				retrievePod: mockValidPod(),
-				fetchMetadataClient: func(config *Config) (metadata.Fetcher, error) {
+				fetchMetadataClient: func(config *Config) (metadata.ReadWriteClient, error) {
 					return testutil.NilFetcher()()
```
aysylu: nit: could we rename this to `NilClient` or `NilReadWriteClient` to be consistent with the new naming?

aysylu:
@nenaddedic confirmed that the integration tests pass locally. The failure of the test is caused by the pod from another test still being in the `Terminating` state, which leads to the unforeseen failure in the test. This is a flake and should be fixed as part of #326

Issue #427, opened 2019-12-18
'signature' field in an Attestation Occurrence should be base64 encoded


nenaddedic:
## Expected Behavior
occurrence.attestation.attestation.pgpSignedAttestation.signature should be base64 encoded.  In particular:
- ISP should create base64 encoded signatures
- both ISP and GAP should expect base64 encoded signatures

This is both consistent with Binary Authorization (see [1]), and easier to use because it is otherwise difficult to deal with proper formatting of PGP armor.

## Actual Behavior
'signature' field is assumed to be PGP armor without any further encoding.

## Steps to Reproduce the Problem
I believe that problems described in [2] are caused by this.

[1] https://cloud.google.com/binary-authorization/docs/making-attestations
[2] https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/kritis-users/atVVPZmzs24/xyb-oKvYAAAJ

Tharun-Sabbu:
Hello, I'm having the exact same issue. I was using standalone setup, can you please release a version with this fix. 

And also I had few other issues with helm chart for kritis (Kritis Helm Chart v0.1.0), I'm on Kubernetes Version 1.15.5 and using helm v2.16.1. Preinstall pod was failing each time with different errors when I try to install chart:

> time="2020-01-03T03:44:53Z" level=info msg="[kubectl apply -f -]"
time="2020-01-03T03:44:53Z" level=info
time="2020-01-03T03:44:53Z" level=error msg="error: SchemaError(io.k8s.api.apps.v1beta2.ReplicaSet): invalid object doesn't have additional properties\n"
time="2020-01-03T03:44:53Z" level=fatal msg="exit status 1"

> time="2020-01-03T06:34:18Z" level=info msg="[kubectl apply -f -]"
time="2020-01-03T06:34:18Z" level=info
time="2020-01-03T06:34:18Z" level=error msg="error: SchemaError(io.k8s.api.apps.v1beta2.ScaleSpec): invalid object doesn't have additional properties\n"
time="2020-01-03T06:34:18Z" level=fatal msg="exit status 1"


To overcome this issue, I had to disable preinstall and postinstall. Then convert Go lang to shell scripts.

ooq:
This issue will be reverted.

Notes on the issue in [2]:
Signature field in attestation is a string type in [v1beta1 API](https://github.com/grafeas/grafeas/blob/master/proto/v1beta1/attestation.proto#L39), so it should not be base64 encoded.
[...]

PR #429, opened 2019-12-20
Base64 signature

nenaddedic:
This is a fix for https://github.com/grafeas/kritis/issues/427.

Changes:
- When verifying an attestation, treat the signature as base64-encoded.
- When creating an attestation, create it with a base64-encoded signature.

This change breaks backwards compatibility.  Attestations verifiable with old code are no longer verifiable with the new code.

Users of GAP will have to recreate the attestations.

Effects on ISP will be limited: attestations previously created by ISP will become invalid, but ISP will create new valid ones for images passing the ISP.  The only tricky case is that of a deployment previously passing the ISP (under the old code), but which is now not passing.  In this case, new Pods will trigger ISP re-evaluation (because invalid attestations, under the new code), which will fail.

### DIFF
--- a/pkg/kritis/review/validating_transport.go
+++ b/pkg/kritis/review/validating_transport.go
@@ -17,6 +17,8 @@ limitations under the License.
 package review
 
 import (
+	"encoding/base64"
+
 	"github.com/golang/glog"
 	"github.com/grafeas/kritis/pkg/kritis/apis/kritis/v1beta1"
 	"github.com/grafeas/kritis/pkg/kritis/attestation"
@@ -41,7 +43,7 @@ func (avt *AttestorValidatingTransport) GetValidatedAttestations(image string) (
 	keys := map[string]string{}
 	key, fingerprint, err := secrets.KeyAndFingerprint(avt.Attestor.Spec.PublicKeyData)
 	if err != nil {
-		glog.Errorf("Error parsing key for %q: %v", avt.Attestor.Name, err)
+		glog.Infof("Error parsing key for %q: %v", avt.Attestor.Name, err)
 		return nil, err
 	}
 	keys[fingerprint] = key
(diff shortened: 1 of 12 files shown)


pkg/kritis/review/validating_transport.go
```diff
@@ -41,7 +43,7 @@ func (avt *AttestorValidatingTransport) GetValidatedAttestations(image string) (
 	keys := map[string]string{}
 	key, fingerprint, err := secrets.KeyAndFingerprint(avt.Attestor.Spec.PublicKeyData)
 	if err != nil {
-		glog.Errorf("Error parsing key for %q: %v", avt.Attestor.Name, err)
+		glog.Infof("Error parsing key for %q: %v", avt.Attestor.Name, err)
```
aysylu: What was the rationale behind downgrading the severity of the log message here?

pkg/kritis/review/validating_transport.go
```diff
@@ -58,11 +60,16 @@ func (avt *AttestorValidatingTransport) GetValidatedAttestations(image string) (
 		return nil, err
 	}
 	for _, a := range attestations {
-		if err = host.VerifyAttestationSignature(keys[a.KeyID], a.Signature); err != nil {
-			glog.Errorf("Could not find or verify attestation for attestor %s: %s", a.KeyID, err.Error())
-		} else {
-			out = append(out, attestation.ValidatedAttestation{AttestorName: avt.Attestor.Name, Image: image})
+		decoded_sig, err := base64.StdEncoding.DecodeString(a.Signature)
+		if err != nil {
+			glog.Infof("Cannot base64 decode signature: %v", err)
```
aysylu: should this be of severity `Error`?

pkg/kritis/review/validating_transport_test.go
```diff
@@ -31,6 +31,10 @@ import (
 	"github.com/grafeas/kritis/pkg/kritis/util"
 )
 
+func enc(in string) string {
```
aysylu: Let's rename this to `to64b` or `encode64b` or something similar, so that it's more readable at the call sites without having to check the definition.

pkg/kritis/review/validating_transport.go
nenaddedic: I don't know :)  Restoring to the way it was.

pkg/kritis/review/validating_transport.go
nenaddedic: I think it shouldn't.  Anyone can create any kind of Occurrence, and there's nothing we can do about it.

pkg/kritis/review/validating_transport_test.go
nenaddedic: Done.

pkg/kritis/review/validating_transport.go
aysylu: good point, sgtm

aysylu:
@nenaddedic: before merging, could you please confirm integration tests pass locally for you? The kokoro results look like a flake, but I'd like to confirm first.

nenaddedic:
> @nenaddedic: before merging, could you please confirm integration tests pass locally for you? The kokoro results look like a flake, but I'd like to confirm first.

@aysylu yes, they pass locally

PR #437, opened 2019-12-30
Add integration test for GAP

nenaddedic:
This PR adds an integration test for GAP.

It expects to find certain Notes and Occurrences in the test project.  To make it easy to create these Notes and Occurrences, the PR contains a script to do so.  Keys used to sign attestation Occurrences are checked in also as part of this PR.  (No security issues here, because it's just keys used for testing.)

Note:  Before merging this, we'll have to create these required Notes and Occurrences in the "kritis-int-test" project.


integration/run_test.go
```diff
@@ -293,18 +300,107 @@ func setUp(t *testing.T) (kubernetes.Interface, *v1.Namespace, func(t *testing.T
...
+	}
+}
+
+func TestKritisGAPLogic(t *testing.T) {
+	cs, ns, tearDown := setUpGAP(t)
+	defer tearDown(t)
+
+	var testCases = []struct {
+		template string
+
```
aysylu: empty line

integration/testdata/generic-attestation-policy/test-attestor-2.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kritis.grafeas.io/v1beta1
```
aysylu: Is this needed? I don't see any references to it in the PR.

Makefile
```diff
@@ -209,6 +215,13 @@ setup-integration-local: setup-integration-local
 	gcloud -q container images add-tag \
 		gcr.io/kritis-tutorial/nginx-no-digest:latest \
 		gcr.io/$(GCP_PROJECT)/nginx-no-digest:latest
+	gcloud projects add-iam-policy-binding ${GCP_PROJECT} \
+		--member=serviceAccount:kritis-ca-admin@${GCP_PROJECT}.iam.gserviceaccount.com \
+		--role=roles/containeranalysis.notes.occurrences.viewer
+	gcloud projects add-iam-policy-binding ${GCP_PROJECT} \
+		--member=serviceAccount:kritis-ca-admin@${GCP_PROJECT}.iam.gserviceaccount.com \
+		--role=roles/containeranalysis.occurrences.viewer
```
ooq: `https://github.com/grafeas/kritis/blob/master/docs/install.md` also lists `editor` permissions, and `notes.viewer`. Not needed?

integration/run_test.go
```diff
@@ -230,8 +230,7 @@ func installKritis(cs kubernetes.Interface, ns *v1.Namespace) (func(*testing.T),
 	return cleanup, nil
 }
 
-// Complete setUp for a test. Returns a tearDown function.
-func setUp(t *testing.T) (kubernetes.Interface, *v1.Namespace, func(t *testing.T)) {
+func setUpKritisInNS(t *testing.T) (kubernetes.Interface, *v1.Namespace, func(*testing.T)) {
```
ooq: `InNS` stands for?

Makefile
```diff
@@ -179,8 +179,11 @@ gcb-signer-push-image: gcb-signer-image
 	docker push $(REGISTRY)/kritis-gcb-signer:$(IMAGE_TAG)
 
 # Fully setup local integration testing: only needs to run just once
+# TODO: move entire setup into bash script
+# TODO: enable necessary dependency APIs (GKE, containeranslysis) or give instructions to do so
```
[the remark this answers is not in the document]
nenaddedic: Yes, API enablement is.

hack/setup-containeranalysis-resources.sh
```diff
@@ -0,0 +1,183 @@
+#!/bin/bash
+
+# Copyright 2018 Google LLC
```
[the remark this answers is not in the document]
nenaddedic: I updated boilerplate.py.  Do I need to update anything else?

integration/testdata/generic-attestation-policy/test-attestor-2.yaml
nenaddedic: I think I'll add another test that uses it.  I can do it in a separate PR if you prefer it, in which case I'll get rid of these unused objects.  Let me know.

Makefile
nenaddedic: Sigh... I think occurrences.editor is kindof needed, because ISP creates attestations.  But this behavior (caching ISP decisions as attestations) is not really exercised in the test, so in reality the permission is not required.  I added it anyway...

I don't think note editing is needed though.

hack/setup-containeranalysis-resources.sh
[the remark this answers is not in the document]
aysylu: Nope, if it passes, then we're all good here.

integration/testdata/generic-attestation-policy/test-attestor-2.yaml
aysylu: Yes please, let's make sure we keep the history as understandable as possible. Thanks!

Makefile
nenaddedic: No, the setup script does that, it's a one-time thing.

integration/testdata/generic-attestation-policy/test-attestor-2.yaml
nenaddedic: Instead of adding another test, I changed TestKritisGAPLogic very slightly.  The policy under test now contains two attestors, to make sure that needs-only-one-attestor semantics is exercised.  And the test now uses test-attestor-2.  WDYT?

aysylu: Perfect, thank you. I like the updated version of the test. One minor optional thing: if you could please comment on top of the test the high-level of what the test is verifying, that'd be fantastic.

nenaddedic: Thanks, done.

PR #438, opened 2020-01-02
Interpret aa's NoteReference as full note name, not just project id

nenaddedic:
This is to address https://github.com/grafeas/kritis/issues/420


docs/resources.md
```diff
@@ -193,7 +193,7 @@ metadata:
     name: qa-attestator
     namespace: qa
 spec:
-    noteReference: projects/image-attestor
+    noteReference: projects/image-attestor/notes/qa-note
```
jialur: Should this be "qa-attestator" instead?

pkg/kritis/metadata/metadata.go
```diff
@@ -68,14 +68,11 @@ type PGPAttestation struct {
 	OccID string
 }
 
-// GetProjectFromNoteReference extracts the project ID form the NoteReference
-func GetProjectFromNoteReference(ref string) (string, error) {
-	str := strings.Split(ref, "/")
-	if len(str) != 2 {
-		return "", fmt.Errorf("invalid Note Reference, should be in format projects/<project_id>")
+// ParseNoteReference extracts the project ID and the note ID from the NoteReference.
```
jialur: Remove the period to be consistent with the other comments in the files.

PR #445, opened 2020-01-07
Supporting Multiple Keys in Kritis AA.

ooq:
This change is part of binauthz compatibility changes.
It will support multiple keys in kritis attestation authority, with “verifiable by ANY key” verification semantics. This is mainly to support key rotation


pkg/kritis/apis/kritis/v1beta1/attestationauthority.go
```diff
@@ -33,10 +33,10 @@ type AttestationAuthority struct {
 
 // AttestationAuthoritySpec is the spec for a AttestationAuthority resource
 type AttestationAuthoritySpec struct {
-	NoteReference        string `json:"noteReference"`
-	PrivateKeySecretName string `json:"privateKeySecretName"`
-	PublicKeyData        string `json:"publicKeyData"`
-	PolicyType           string `json:"policyType"`
+	NoteReference        string   `json:"noteReference"`
+	PrivateKeySecretName string   `json:"privateKeySecretName"`
+	PublicKeyList        []string `json:"publicKeyList"`
```
nenaddedic: Do you think we can immediately change this to a key wrapper like [1]? (We'll probably want to add PKIX support soon enough, so might as well prepare for that.)

[1] https://cloud.google.com/binary-authorization/docs/reference/rest/v1/projects.attestors#AttestorPublicKey

pkg/kritis/review/validating_transport.go
```diff
@@ -41,12 +42,18 @@ type AttestorValidatingTransport struct {
...
-	key, fingerprint, err := secrets.KeyAndFingerprint(avt.Attestor.Spec.PublicKeyData)
-	if err != nil {
-		glog.Errorf("Error parsing key for %q: %v", avt.Attestor.Name, err)
-		return nil, err
+	for _, keyData := range avt.Attestor.Spec.PublicKeyList {
+		key, fingerprint, err := secrets.KeyAndFingerprint(keyData)
+		if err != nil {
+			glog.Warningf("Error parsing key for %q: %v", avt.Attestor.Name, err)
+		} else {
+			keys[fingerprint] = key
```
nenaddedic: Also log warning if duplicate found.

pkg/kritis/review/validating_transport.go
```diff
@@ -41,12 +42,18 @@ type AttestorValidatingTransport struct {
...
+		}
+	}
+	if len(keys) == 0 {
+		glog.Errorf("Unable to find any valid key for %q", avt.Attestor.Name)
```
nenaddedic: Do we need to log error message and return it as well?

pkg/kritis/review/validating_transport_test.go
```diff
@@ -50,14 +50,21 @@ func TestValidatingTransport(t *testing.T) {
 		ObjectMeta: metav1.ObjectMeta{Name: "test-attestor"},
 		Spec: v1beta1.AttestationAuthoritySpec{
 			PrivateKeySecretName: "test-success",
-			PublicKeyData:        base64.StdEncoding.EncodeToString([]byte(pub)),
+			PublicKeyList:        []string{base64.StdEncoding.EncodeToString([]byte(pub))},
+		},
+	}
+	validAuthWithTwoKeys := v1beta1.AttestationAuthority{
```
nenaddedic: Could you add some tests where the auth has two valid keys?  Mainly to exercise the disjunctive logic (need attestation by only one key).  Thanks!

pkg/kritis/apis/kritis/v1beta1/attestationauthority.go
ooq: I think we should probably address this in a separate PR. I documented it in #453 .
WDYT?

pkg/kritis/review/validating_transport_test.go
ooq: Done.

pkg/kritis/review/validating_transport.go
ooq: Removed logging (so it will instead be logged once at higher level.)

pkg/kritis/review/validating_transport.go
ooq: Done.

pkg/kritis/apis/kritis/v1beta1/attestationauthority.go
nenaddedic: sgtm

PR #446, opened 2020-01-08
Move private secret key out of AttestationAuthority.

ooq:
This change is part of binauthz compatibility changes.
It will move private secret key name out of kritis attestation authority, and into the policies, i.e., the ISP policy and the build policy. 
The purpose of this is to separate the semantics of `signing` and `verifying`, leaving attestation authority to do only `verifying` and the respective policies to do `signing`.

### DIFF
--- a/integration/crd.go
+++ b/integration/crd.go
@@ -26,7 +26,6 @@ import (
 	"time"
 
 	integration_util "github.com/grafeas/kritis/pkg/kritis/integration_util"
-	"github.com/grafeas/kritis/pkg/kritis/testutil"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
@@ -49,13 +48,8 @@ var crdNames = map[string]string{
 	"imagesecuritypolicies.kritis.grafeas.io": "my-isp",
 }
 
-func createAttestationAuthority(t *testing.T, project string, ns string) {
+func createKeySecret(t *testing.T, project string, ns string, pubKey string, privKey string) {
 	t.Helper()
-	// Generate a key value pair
-	pubKey, privKey := testutil.CreateKeyPair(t, aaSecret)
-	// get the base encoded value for public key.
-	pubKeyEnc := base64.StdEncoding.EncodeToString([]byte(pubKey))
-
 	// create a tmp dir for keys
 	d, err := ioutil.TempDir("", "_keys")
 	if err != nil {
@@ -76,9 +70,6 @@ func createAttestationAuthority(t *testing.T, project string, ns string) {
 	if _, err := integration_util.RunCmdOut(cmd); err != nil {
 		t.Fatalf("unexpected error %s", err)
 	}
-
-	// Create the Attestation authority
-	createAA(t, project, ns, pubKeyEnc)
 }
 
 func waitForCRDExamples(t *testing.T, ns *v1.Namespace) {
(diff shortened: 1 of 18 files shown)


pkg/kritis/apis/kritis/v1beta1/securitypolicy.go
```diff
@@ -36,6 +36,7 @@ type ImageSecurityPolicySpec struct {
 	ImageAllowlist                   []string                         `json:"imageAllowlist"`
 	PackageVulnerabilityRequirements PackageVulnerabilityRequirements `json:"packageVulnerabilityRequirements"`
 	AttestationAuthorityName         string                           `json:"attestationAuthorityName"`
```
nenaddedic: Let's make sure to also change this to Note name.  (Same in build policy.)  We can, and maybe should, do it in a separate PR, let's just not forget about it.

pkg/kritis/gcbsigner/signer.go
```diff
@@ -66,14 +66,14 @@ func (s Signer) ValidateAndSign(prov BuildProvenance, bps []v1beta1.BuildPolicy)
...
-		if err := s.addAttestation(prov.ImageRef, bp.Namespace, bp.Spec.AttestationAuthorityName); err != nil {
+		if err := s.addAttestation(prov.ImageRef, bp.Namespace, bp.Spec.AttestationAuthorityName, bp.Spec.PrivateKeySecretName); err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
-func (s Signer) addAttestation(image string, ns string, authority string) error {
+func (s Signer) addAttestation(image string, ns string, authority string, keyName string) error {
```
nenaddedic: Maybe call it keySecretName or something to make it clear this is a k8s secret name, not key id.

integration/crd.go
```diff
@@ -38,11 +38,10 @@ metadata:
   name: test-attestor
 spec:
   noteReference: projects/%s
-  privateKeySecretName: %s
   publicKeyData: %s`
 )
 
-// Secret name for test-attestor
+// Secret name for test-attestor, which matches integration/testdata/image-security-policy/my-isp.yaml
```
nenaddedic: Please consider moving the key pair generation out of createAA, because AA now only has the public portion.  One way to do it:
- move createAA call out of here and into setupISP() in integration/run_test.go
- rename this to createSecret or something like it
  + and change this to return the generated public key

WDYT?

pkg/kritis/apis/kritis/v1beta1/securitypolicy.go
ooq: Ack.

integration/crd.go
ooq: Done.

pkg/kritis/gcbsigner/signer.go
ooq: Done.

integration/run_test.go
```diff
@@ -278,7 +278,13 @@ func setUp(t *testing.T) (kubernetes.Interface, *v1.Namespace, func(t *testing.T
 		t.Fatalf("install: %v", err)
 	}
 
-	createAttestationAuthority(t, *gcpProject, ns.Name)
+	// Generate a key value pair
+	pubKey, privKey := testutil.CreateKeyPair(t, aaSecret)
```
nenaddedic: I think 'aaSecret' can go.  Just write 'test-attestor' here and comment that it must be the same as in my-isp.yaml.

ooq: It's also used in "cleanup.go". so it might still be easier to centralize it?

nenaddedic: Oh, OK, I didn't know that.  Then let's leave it.

PR #449, opened 2020-01-09
Change ANY attestor to ALL attestor semantic in GAP.

ooq:
This PR will change Kritis Generic Attestation Policy (GAP) semantic to “all attestors”. Currently the semantic is ANY for the purpose of key rotation, which will be addressed by the previous proposed change (supporting multiple keys in Kritis attestor).  

Also edge case: if a GAP has no attestor, the image will still be admitted.

If there are multiple GAPs, only one GAP needs to be satisfied for image to be admitted.

The change  is in part to get compatibility with Binary Authorization.

### DIFF
--- a/docs/resources.md
+++ b/docs/resources.md
@@ -90,7 +90,7 @@ Generic Attestation Policy Spec description:
 | attestationAuthorityNames | | Non-empty List of [Attestation Authorities](#attestationauthority-crd) for which all of them are required to be satisfied before the Admission Controller will admit the pod.|
 | attestationAuthorityNames | | List of [Attestation Authorities](#attestationauthority-crd) for which one of is required to be satisfied before the Admission Controller will admit the pod.|
 
-Note that the list of [Attestation Authorities](#attestationauthority-crd) must be non-empty. If the list is empty, images can only be admitted through allowlist (GAP allowlist is WIP).
+Note that the list of [Attestation Authorities](#attestationauthority-crd) must be non-empty. If the list is empty, an error will be thrown for malformed policy, and no image will be admitted, including allowlisted images.
 
 Admission Allowlist Pattern Spec description
 
--- a/docs/resources.md
+++ b/docs/resources.md
@@ -36,7 +36,6 @@ kubetl get pods -l kritis.grafeas.io/invalidImageSecPolicy=invalidImageSecPolicy
 
 GenericAttestationPolicy (GAP) is a Custom Resource Definition which enforces policies based on pre-existing attestations.
 The policy expects ALL attestation authorities to be satisfied before allowing the container image to be admitted.
-If no attestation authority is present in GAP, the container image will be admitted automatically.
 As opposed to [ISPs](#imagesecuritypolicy-crd) the GAP does not create new attestations.
 The general use case for GAPs are to have a policy that enforces attestations that have come from your CI pipeline, or other places in your release pipeline.
 
@@ -85,7 +84,9 @@ Generic Attestation Policy Spec description:
 
 | Field     | Default (if applicable)   | Description |
 |-----------|---------------------------|-------------|
-| attestationAuthorityNames | | List of [Attestation Authorities](#attestationauthority-crd) for which one of is required to be satisfied before the Admission Controller will admit the pod.|
+| attestationAuthorityNames | | Non-empty List of [Attestation Authorities](#attestationauthority-crd) for which all of them are required to be satisfied before the Admission Controller will admit the pod.|
+
+Note that the list of [Attestation Authorities](#attestationauthority-crd) must be non-empty. If the list is empty, images can only be admitted through allowlist (GAP allowlist is WIP).
 
 ## ImageSecurityPolicy CRD
 
--- a/docs/resources.md
+++ b/docs/resources.md
@@ -35,7 +35,8 @@ kubetl get pods -l kritis.grafeas.io/invalidImageSecPolicy=invalidImageSecPolicy
 ## GenericAttestationPolicy CRD
 
 GenericAttestationPolicy (GAP) is a Custom Resource Definition which enforces policies based on pre-existing attestations.
-The policy expects any one attestation authority to be satisfied before allowing the container image to be admitted.
+The policy expects ALL attestation authorities to be satisfied before allowing the container image to be admitted.
+If no attestation authority is present in GAP, the container image will be admitted automatically.
 As opposed to [ISPs](#imagesecuritypolicy-crd) the GAP does not create new attestations.
 The general use case for GAPs are to have a policy that enforces attestations that have come from your CI pipeline, or other places in your release pipeline.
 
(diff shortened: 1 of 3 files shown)


pkg/kritis/review/review.go
```diff
@@ -80,24 +80,24 @@ func (r Reviewer) ReviewGAP(images []string, gaps []v1beta1.GenericAttestationPo
...
 			}
 		}
-		if err := r.config.Strategy.HandleAttestation(image, pod, imgAttested); err != nil {
+		if err := r.config.Strategy.HandleAttestation(image, pod, imgAttestedByAll); err != nil {
 			glog.Errorf("error handling attestations %v", err)
 		}
-		if !imgAttested {
-			return fmt.Errorf("image %s is not attested", image)
+		if !imgAttestedByAll {
+			return fmt.Errorf("image %s is not attested by all authories in any gap policy", image)
```
nenaddedic: It would be better if the error message named all the offending images, and for each image the unsatisfied auth.  Something like this:
- nginx@sha256:abcd not attested by auth1, auth2
- bla@sha256:1234 not attested by auth1

docs/resources.md
```diff
@@ -35,7 +35,8 @@ kubetl get pods -l kritis.grafeas.io/invalidImageSecPolicy=invalidImageSecPolicy
 ## GenericAttestationPolicy CRD
 
 GenericAttestationPolicy (GAP) is a Custom Resource Definition which enforces policies based on pre-existing attestations.
-The policy expects any one attestation authority to be satisfied before allowing the container image to be admitted.
+The policy expects ALL attestation authorities to be satisfied before allowing the container image to be admitted.
+If no attestation authority is present in GAP, the container image will be admitted automatically.
```
nenaddedic: So this one is slightly tricky.  I think we should respect the semantics of [1] as in REQUIRE_ATTESTATION mode.  So that means that no aas is an error, rather than "admit".

[1] https://cloud.google.com/binary-authorization/docs/reference/rest/v1/Policy#AdmissionRule

pkg/kritis/review/review_test.go
```diff
@@ -123,80 +132,88 @@ func TestReviewGAP(t *testing.T) {
...
 			shouldErr:    false,
 		},
 		{
 			name:         "image attested by one attestor out of two",
 			image:        img,
 			policies:     gapWithTwoAAs,
 			attestations: validAtts,
-			isAttested:   true,
-			shouldErr:    false,
+			isAdmitted:   false,
```
nenaddedic: Let's also add the positive case (attested by two out of two, therefore admitted).

pkg/kritis/review/review.go
ooq: Done

docs/resources.md
```diff
@@ -85,7 +84,9 @@ Generic Attestation Policy Spec description:
 
 | Field     | Default (if applicable)   | Description |
 |-----------|---------------------------|-------------|
-| attestationAuthorityNames | | List of [Attestation Authorities](#attestationauthority-crd) for which one of is required to be satisfied before the Admission Controller will admit the pod.|
+| attestationAuthorityNames | | Non-empty List of [Attestation Authorities](#attestationauthority-crd) for which all of them are required to be satisfied before the Admission Controller will admit the pod.|
+
+Note that the list of [Attestation Authorities](#attestationauthority-crd) must be non-empty. If the list is empty, images can only be admitted through allowlist (GAP allowlist is WIP).
```
nenaddedic: Mention that evaluating a policy with an empty aa list results in an error.

nenaddedic: Come to think of it, we should both understand and document the error behavior.  (Probably depends on the webhook config.)

Nothing to do with this PR though...

pkg/kritis/review/review.go
```diff
@@ -61,8 +61,35 @@ func generateGapAllowlist(gaps []v1beta1.GenericAttestationPolicy) []string {
 	return allowlist
 }
 
+// ReviewImageWithGAP reviews single image against a single generic attestation policy
+func (r Reviewer) ReviewImageWithGAP(image string, gap v1beta1.GenericAttestationPolicy, c metadata.ReadOnlyClient) (bool, []string, error) {
+	glog.Infof("Validating against GenericAttestationPolicy %s", gap.Name)
+
+	// Get all AttestationAuthorities in this policy.
```
nenaddedic: This whole section should be in ReviewGAP.  Then you can error out immediately (before even attempting to review any images) if there are no auths.  (Also no need to call getAttestationAuthoritiesForGAP() for every image.)

pkg/kritis/review/review.go
```diff
@@ -61,8 +61,35 @@ func generateGapAllowlist(gaps []v1beta1.GenericAttestationPolicy) []string {
...
+	}
+	glog.Infof("Image is missing attestations from %+q in gap %s", notAttestedAuthNames, gap.Name)
+
+	return false, notAttestedAuthNames, nil
+}
+
 // ReviewGAP reviews images against generic attestation policies
-// Returns error if violations are found and handles them per violation strategy
 func (r Reviewer) ReviewGAP(images []string, gaps []v1beta1.GenericAttestationPolicy, pod *v1.Pod, c metadata.ReadOnlyClient) error {
 	images = util.RemoveGloballyAllowedImages(images)
```
nenaddedic: I would do the "are there any aas" check first, and error out if not.  IMO, this will result in less surprising behavior.  (Because right now, whether a policy error is returned depends on the images being reviewed, not only on the policy itself.)

docs/resources.md
ooq: I thought about saying "results in an error", then I thought it would exclude the case for allowlisted images + 0 AA gap. Be more explicit for the non-allowlisted case?

What other behaviors and interactions with webhook config are you thinking?

nenaddedic: PTAL at my comment in review.go.  I think that "0 AAs in GAP" should always be an error, regardless of the request being checked.

pkg/kritis/review/review.go
ooq: Per discussion, adding `checkGAPs()` in the beginning of `ReviewGAP()` to throw out error early. Minor: now it's still calling `getAttestationAuthoritiesForGAP()` for every image.

docs/resources.md
ooq: Done.

pkg/kritis/review/review_test.go
```diff
@@ -91,6 +105,18 @@ func TestReviewGAP(t *testing.T) {
 			AttestationAuthorityNames: []string{"test", "test2"},
 		},
 	}}
+	// One policy without attestor:
```
nenaddedic: I think L99 above (comment explaining gapWithTwoAAs) is now incorrect.  There is an image which satisfies both auths.

pkg/kritis/review/review_test.go
```diff
@@ -121,79 +147,103 @@ func TestReviewGAP(t *testing.T) {
...
 			attestations: []metadata.PGPAttestation{},
-			isAttested:   false,
+			isAdmitted:   false,
 			shouldErr:    false,
 		},
 		{
 			name:         "image attested by one attestor out of two",
 			image:        img,
 			policies:     gapWithTwoAAs,
 			attestations: validAtts,
```
nenaddedic: Can you rename "validAtts" to something more easily distinguishable from "twoValidAtts"?

PR #456, opened 2020-01-15
Remove artifacts/integration-examples/image-security-policy-example.yaml

ooq:
This yaml is not used anywhere. Looking at file history, it was probably used in ancient time, until all integration testing-related files were moved to /integration.


aysylu:
Thanks for the cleanup!

PR #457, opened 2020-01-21
Updated cron.go

judavi:
Fixed small typo. It should pod instead of po


ooq:
Thanks for the fix @judavi .
It might just be k8s standard abbreviation, as in pod (po), service (svc), deployment (deploy), etc. 
Let's also make it consistent with https://github.com/grafeas/kritis/blob/71a1d2c56f1549e2e15512058e961ab932b8de85/docs/tutorial.md#L322

judavi:
Sure @ooq !

aysylu:
Good catch, @judavi! Thanks for the fix.

Issue #458, opened 2020-01-21
ImageSecurityPolicy Validation


judavi:
## Expected Behavior
The ImageSecurityPolicy should block any new pod that has vulnerabilities reported on Grafeas

## Actual Behavior
I'm running the standalone example, but I'm enabling exclusively the ImageSecurityPolicy 
```
➜  kubectl describe ImageSecurityPolicy my-isp

Name:         my-isp
Namespace:    default
Labels:       <none>
Annotations:  kubectl.kubernetes.io/last-applied-configuration:
                {"apiVersion":"kritis.grafeas.io/v1beta1","kind":"ImageSecurityPolicy","metadata":{"annotations":{},"name":"my-isp","namespace":"default"}...
API Version:  kritis.grafeas.io/v1beta1
Kind:         ImageSecurityPolicy
Metadata:
  Creation Timestamp:  2020-01-21T14:46:27Z
  Generation:          1
  Resource Version:    17991
  Self Link:           /apis/kritis.grafeas.io/v1beta1/namespaces/default/imagesecuritypolicies/my-isp
  UID:                 cdadd1b9-3c5c-11ea-bd76-6e9893f6ce69
Spec:
  Package Vulnerability Requirements:
    Maximum Severity:  BLOCK_ALL
Events:                <none>
```
My information in Grafeas looks like, so I have occurrences for the sha256 and for the tag (just in case, I'm still experimenting) 
```
GET https://grafeas-server:443/v1beta1/projects/kritis/occurrences

{
    "occurrences": [
        {
            "name": "projects/kritis/occurrences/40b970ce-98a9-4423-a72a-8974a9b3534e",
            "resource": {
                "name": "",
                "uri": "gcr.io/kritis-tutorial/java-with-vulnz@sha256:358687cfd3ec8e1dfeb2bf51b5110e4e16f6df71f64fba01986f720b2fcba68a",
                "contentHash": null
            },
            "noteName": "projects/kritis/notes/CVE-2015-2730",
            "kind": "VULNERABILITY",
            "remediation": "",
            "createTime": "2020-01-21T14:30:15.437255449Z",
            "updateTime": "2020-01-21T14:30:15.437255449Z",
            "vulnerability": {
                "type": "",
                "severity": "MEDIUM",
                "cvssScore": 4.3,
                "packageIssue": [
                    {
                        "affectedLocation": {
                            "cpeUri": "cpe:/o:redhat:enterprise_linux:7",
                            "package": "imagemagick",
[...]

PR #460, opened 2020-01-23
Update securitypolicy.go

judavi:
Following the convention of the other fields, this field should be maximumFixUnavailableSeverity. Also following the documentation from here [resources.md](https://github.com/grafeas/kritis/blob/71a1d2c56f1549e2e15512058e961ab932b8de85/docs/resources.md) the field should be maximumFixUnavailableSeverity


judavi:
This difference on the name field is related to https://github.com/grafeas/kritis/issues/458

ooq:
A good to have: Do you think we can add some integration tests to catch this? Thanks.

aysylu:
@judavi: could you please clarify if this is fixing an existing bug or is a documentation/naming consistency fix?

judavi:
@aysylu I'll say both?
- Documentation consistency :
https://github.com/grafeas/kritis/blob/71a1d2c56f1549e2e15512058e961ab932b8de85/docs/resources.md
Captura de Pantalla 2020-01-30 a la(s) 7 57 58 a  m

- And struct consistency/bug? because in the structs the all the fields/values has the same name in camel case, but this one, in particular, is not following the convention:
Captura de Pantalla 2020-01-30 a la(s) 7 59 34 a  m

So because of that, I was struggling a little bit to verify if the image vulnerabilities policy was working correctly or not.

By the way @ooq I'll be happy to add the integration test. But, could you guide me where these tests should be added? and a brief idea of how what we should be testing? I mean, name conventions? or a particular test for MaximumFixUnavailableSeverity?

ooq:
Hi @judavi , thanks for the explanation!
The integration tests are all here: https://github.com/grafeas/kritis/blob/master/integration/run_test.go
The existing tests unfortunately does not test this field, so it will be good to add that.

It is strange given that the ISP yaml in integration test is already using the correct field name. 
https://github.com/grafeas/kritis/blob/6b52ba2b6aa78b8fe087df8c8c16fa667405f0f9/integration/testdata/image-security-policy/my-isp.yaml#L13

ooq:
@aysylu Do you know why kokoro test is not triggered on this PR?

aysylu:
@judavi: thanks for clarifying! Sounds good and we appreciate your thorough investigation and the fix in this PR. I'll let @ooq guide you through adding the integration tests. Thanks for your active contribution to Grafeas and Kritis!

aysylu:
@ooq: for posterity, you'll want to add `kokoro:run` label to run integration tests.

ooq:
Update: i'll merge this PR once tests pass.

ba32107:
Hello, we just ran into this problem ourselves and it took us quite a while to investigate the root cause. Just bumping on this PR to have it merged, because I believe currently, if someone sets up an `ImageSecurityPolicy` according to the docs, it will simply not work due to this issue.

ooq:
@ba32107 This PR was approved but pending on some test issues. Let me see if I can fix it and get it merged in. Thanks for the feedback!

ooq:
It looks like [image-security-policy-crd.yaml](https://github.com/grafeas/kritis/blob/faeba81c520ca742fe378cc87d99e04bdda098c1/artifacts/image-security-policy-crd.yaml#L26) would also needs to be updated accordingly.

ooq:
Merged. Thanks for the PR @judavi !

PR #461, opened 2020-01-28
Updated README

aysylu:
- added instructions for standalone Kritis
- cleaned up obsolete references

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,11 +38,11 @@ Authorization](https://cloud.google.com/binary-authorization/).  For details see
 
 ## Getting Started
 
-* Watch the talk on [Software Supply Chain Management with Grafeas and Kritis](https://www.infoq.com/presentations/supply-grafeas-kritis/)
-* Learn the concepts in the [Kritis whitepaper](docs/binary-authorization.md)
-* Try out Kritis with standalone Grafeas by following [instructions](docs/standalone/README.md)
-* Try the [Tutorial](docs/tutorial.md) to learn how to block vulnerabilities using [GCP Container Analysis](https://cloud.google.com/container-registry/docs/get-image-vulnerabilities)
-* Read the [Resource Reference](docs/resources.md) to configure and interact with Kritis resources
+* Watch the talk on [Software Supply Chain Management with Grafeas and Kritis](https://www.infoq.com/presentations/supply-grafeas-kritis/).
+* Learn the concepts in the [Kritis Whitepaper](docs/binary-authorization.md).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,10 +40,9 @@ Authorization](https://cloud.google.com/binary-authorization/).  For details see
 
 * Watch the talk on [Software Supply Chain Management with Grafeas and Kritis](https://www.infoq.com/presentations/supply-grafeas-kritis/)
 * Learn the concepts in the [Kritis whitepaper](docs/binary-authorization.md)
-* Get Kritis running with the [Installation guide](docs/install.md)
-* Try the [Tutorial](docs/tutorial.md) to learn how to block vulnerabilities
+* Try out Kritis with standalone Grafeas by following [instructions](docs/standalone/README.md)
```
ooq: Are we suggesting this to be the first thing to try out now?

README.md
```diff
@@ -40,10 +40,9 @@ Authorization](https://cloud.google.com/binary-authorization/).  For details see
 
 * Watch the talk on [Software Supply Chain Management with Grafeas and Kritis](https://www.infoq.com/presentations/supply-grafeas-kritis/)
 * Learn the concepts in the [Kritis whitepaper](docs/binary-authorization.md)
```
ooq: Minor: these are all full sentences, maybe add "." at the end of each.

README.md
ooq: I would expand [instructions] to be more specific, e.g., [standalone mode instructions].

README.md
```diff
@@ -40,10 +40,9 @@ Authorization](https://cloud.google.com/binary-authorization/).  For details see
...
-* Get Kritis running with the [Installation guide](docs/install.md)
-* Try the [Tutorial](docs/tutorial.md) to learn how to block vulnerabilities
+* Try out Kritis with standalone Grafeas by following [instructions](docs/standalone/README.md)
+* Try the [Tutorial](docs/tutorial.md) to learn how to block vulnerabilities using [GCP Container Analysis](https://cloud.google.com/container-registry/docs/get-image-vulnerabilities)
```
ooq: [minor] Similarly for [tutorial]. Also maybe we should just stick to either instruction or tutorial, but not both.

README.md
aysylu: Yes, because that moves us closer to hybrid-cloud solution, which is one of the primary goals of the project.

README.md
aysylu: great point, done

Issue #419, opened 2019-11-04
Figure out client correctly goes through all pages in ListNoteOccurrences call


aysylu:
## Expected Behavior
The client iterates over all pages using  [next_page_token](https://github.com/grafeas/grafeas/blob/master/proto/v1beta1/grafeas.proto#L334)

## Actual Behavior
The client code just returns an iterator: [link](https://godoc.org/cloud.google.com/go/containeranalysis/apiv1beta1#GrafeasV1Beta1Client.ListNoteOccurrences).

## Additional info
This issue is to verify the pagination is handled correctly on the client.

ooq:
I think this is fixed. Closing.

PR #469, opened 2020-02-14
Add test for paginating through Container Analysis vulnerabilities

acamadeo:
API calls to Container Analysis and Grafeas involve pagination when extracting Vulnerabilities and Attestations. This change ensures that calls made to Container Analysis to extract Vulnerabilities properly iterate through all pages.

TESTED= new integration test
ISSUE= #419

### DIFF
--- a/pkg/kritis/metadata/containeranalysis/containeranalysis.go
+++ b/pkg/kritis/metadata/containeranalysis/containeranalysis.go
@@ -41,10 +41,23 @@ const (
 	AttestationAuthority = "ATTESTATION_AUTHORITY"
 )
 
+// For testing -- injectable functions
+var (
+	createListOccurrencesRequest = defaultListOccurrencesRequest
+)
+
 // Client struct implements ReadWriteClient and ReadOnlyClient interfaces.
 type Client struct {
 	client *ca.GrafeasV1Beta1Client
 	ctx    context.Context
+}
+
+func defaultListOccurrencesRequest(containerImage, kind string) *grafeas.ListOccurrencesRequest {
+	return &grafeas.ListOccurrencesRequest{
+		Filter:   fmt.Sprintf("resourceUrl=%q AND kind=%q", util.GetResourceURL(containerImage), kind),
+		Parent:   fmt.Sprintf("projects/%s", getProjectFromContainerImage(containerImage)),
+		PageSize: constants.PageSize,
+	}
 }
 
 // TODO: separate constructor methods for r/w and r/o clients
@@ -107,23 +120,20 @@ func (c Client) fetchVulnerabilityOccurrence(containerImage string, kind string)
 		return nil, fmt.Errorf("%s is not a valid image hosted in GCR", containerImage)
 	}
 
-	req := &grafeas.ListOccurrencesRequest{
-		Filter:   fmt.Sprintf("resourceUrl=%q AND kind=%q", util.GetResourceURL(containerImage), kind),
-		PageSize: constants.PageSize,
-		Parent:   fmt.Sprintf("projects/%s", getProjectFromContainerImage(containerImage)),
-	}
+	req := createListOccurrencesRequest(containerImage, kind)
+	fmt.Println(req)
 
 	it := c.client.ListOccurrences(c.ctx, req)
+	p := iterator.NewPager(it, int(req.PageSize), "")
 	occs := []*grafeas.Occurrence{}
 	for {
-		occ, err := it.Next()
-		if err == iterator.Done {
-			break
-		}
+		nextPageToken, err := p.NextPage(&occs)
 		if err != nil {
 			return nil, err
 		}
-		occs = append(occs, occ)
+		if nextPageToken == "" {
+			break
+		}
 	}
 	return occs, nil
 }
@@ -137,11 +147,10 @@ func (c Client) fetchAttestationOccurrence(containerImage string, kind string, a
 	req := &grafeas.ListNoteOccurrencesRequest{
 		Name: auth.Spec.NoteReference,
 		// Example:
-		// 		Filter:  fmt.Sprintf("resourceUrl=%q AND kind=%q", util.GetResourceURL(containerImage), kind),
+		//              Filter:  fmt.Sprintf("resourceUrl=%q AND kind=%q", util.GetResourceURL(containerImage), kind),
 		Filter:   fmt.Sprintf("resourceUrl=%q", util.GetResourceURL(containerImage)),
 		PageSize: constants.PageSize,
 	}
-
 	occs := []*grafeas.Occurrence{}
 	it := c.client.ListNoteOccurrences(c.ctx, req)
 	for {
--- a/pkg/kritis/metadata/containeranalysis/containeranalysis.go
+++ b/pkg/kritis/metadata/containeranalysis/containeranalysis.go
@@ -121,7 +121,6 @@ func (c Client) fetchVulnerabilityOccurrence(containerImage string, kind string)
 	}
 
 	req := createListOccurrencesRequest(containerImage, kind)
-	fmt.Println(req)
 
 	it := c.client.ListOccurrences(c.ctx, req)
 	p := iterator.NewPager(it, int(req.PageSize), "")
--- a/pkg/kritis/metadata/containeranalysis/containeranalysis.go
+++ b/pkg/kritis/metadata/containeranalysis/containeranalysis.go
@@ -146,7 +146,7 @@ func (c Client) fetchAttestationOccurrence(containerImage string, kind string, a
 	req := &grafeas.ListNoteOccurrencesRequest{
 		Name: auth.Spec.NoteReference,
 		// Example:
-		//              Filter:  fmt.Sprintf("resourceUrl=%q AND kind=%q", util.GetResourceURL(containerImage), kind),
+		// 		Filter:  fmt.Sprintf("resourceUrl=%q AND kind=%q", util.GetResourceURL(containerImage), kind),
 		Filter:   fmt.Sprintf("resourceUrl=%q", util.GetResourceURL(containerImage)),
 		PageSize: constants.PageSize,
 	}
--- a/pkg/kritis/metadata/containeranalysis/containeranalysis.go
+++ b/pkg/kritis/metadata/containeranalysis/containeranalysis.go
@@ -123,16 +123,16 @@ func (c Client) fetchVulnerabilityOccurrence(containerImage string, kind string)
 	req := createListOccurrencesRequest(containerImage, kind)
 
 	it := c.client.ListOccurrences(c.ctx, req)
-	p := iterator.NewPager(it, int(req.PageSize), "")
 	occs := []*grafeas.Occurrence{}
 	for {
-		nextPageToken, err := p.NextPage(&occs)
+		occ, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
 		if err != nil {
 			return nil, err
 		}
-		if nextPageToken == "" {
-			break
-		}
+		occs = append(occs, occ)
 	}
 	return occs, nil
 }
--- a/pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
+++ b/pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
@@ -156,7 +156,7 @@ func TestGetMultiplePagesVulnerabilities(t *testing.T) {
 	}
 
 	if len(vuln) <= 900 {
-		t.Fatalf("Pagination error: expected at least 900 results on image 'gcr.io/kritis-int-test/java-with-vulnz'. Received %client.", len(vuln))
+		t.Fatalf("Pagination error: expected at least 900 results on image 'gcr.io/kritis-int-test/java-with-vulnz'. Received %d.", len(vuln))
 	}
 }
 
--- a/pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
+++ b/pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
@@ -141,7 +141,7 @@ func TestCreateAttestationNoteAndOccurrence(t *testing.T) {
 	}
 }
 
-func TestGetMultiplePages_Vulnerabilities(t *testing.T) {
+func TestGetMultiplePagesVulnerabilities(t *testing.T) {
 	d, err := New()
 	if err != nil {
 		t.Fatalf("Could not initialize the client %s", err)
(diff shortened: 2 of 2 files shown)


pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
```diff
@@ -137,3 +140,30 @@ func TestCreateAttestationNoteAndOccurrence(t *testing.T) {
...
+func TestGetMultiplePages_Vulnerabilities(t *testing.T) {
+	d, err := New()
+	if err != nil {
+		t.Fatalf("Could not initialize the client %s", err)
+	}
+
+	// Set PageSize to 300
+	createListOccurrencesRequest = createListOccurrencesRequest_test
+
+	vuln, err := d.Vulnerabilities("gcr.io/kritis-int-test/java-with-vulnz@sha256:358687cfd3ec8e1dfeb2bf51b5110e4e16f6df71f64fba01986f720b2fcba68a")
```
ooq: How many vulz counts is this image? can we make sure to have 3+ pages?

pkg/kritis/metadata/containeranalysis/containeranalysis.go
```diff
@@ -107,23 +120,20 @@ func (c Client) fetchVulnerabilityOccurrence(containerImage string, kind string)
 		return nil, fmt.Errorf("%s is not a valid image hosted in GCR", containerImage)
 	}
 
-	req := &grafeas.ListOccurrencesRequest{
-		Filter:   fmt.Sprintf("resourceUrl=%q AND kind=%q", util.GetResourceURL(containerImage), kind),
-		PageSize: constants.PageSize,
-		Parent:   fmt.Sprintf("projects/%s", getProjectFromContainerImage(containerImage)),
-	}
+	req := createListOccurrencesRequest(containerImage, kind)
+	fmt.Println(req)
```
aysylu: Please remove debugging statement

pkg/kritis/metadata/containeranalysis/containeranalysis.go
```diff
@@ -137,11 +147,10 @@ func (c Client) fetchAttestationOccurrence(containerImage string, kind string, a
 	req := &grafeas.ListNoteOccurrencesRequest{
 		Name: auth.Spec.NoteReference,
 		// Example:
-		// 		Filter:  fmt.Sprintf("resourceUrl=%q AND kind=%q", util.GetResourceURL(containerImage), kind),
```
aysylu: Why the formatting change?

pkg/kritis/metadata/containeranalysis/containeranalysis.go
```diff
@@ -107,23 +120,20 @@ func (c Client) fetchVulnerabilityOccurrence(containerImage string, kind string)
...
-	req := &grafeas.ListOccurrencesRequest{
-		Filter:   fmt.Sprintf("resourceUrl=%q AND kind=%q", util.GetResourceURL(containerImage), kind),
-		PageSize: constants.PageSize,
-		Parent:   fmt.Sprintf("projects/%s", getProjectFromContainerImage(containerImage)),
-	}
+	req := createListOccurrencesRequest(containerImage, kind)
+	fmt.Println(req)
 
 	it := c.client.ListOccurrences(c.ctx, req)
+	p := iterator.NewPager(it, int(req.PageSize), "")
```
aysylu: What's the reason for doing this differently that in the internal implementation?

pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
```diff
@@ -126,7 +129,7 @@ func TestCreateAttestationNoteAndOccurrence(t *testing.T) {
 		case <-timeout:
 			t.Fatal("Should have created at least 1 occurrence")
 
-			// Got a tick, we should check note occurrences
```
aysylu: revert formatting?

pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
```diff
@@ -137,3 +140,30 @@ func TestCreateAttestationNoteAndOccurrence(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Found err %s", err)
+	}
+
+	if len(vuln) <= 300 {
+		t.Errorf("Pagination error: only received results from first page.")
+	}
+}
+
+func createListOccurrencesRequest_test(containerImage, kind string) *grafeas.ListOccurrencesRequest {
```
aysylu: Let's keep the naming strategy consistent wrt camelCase vs snake_case

pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
```diff
@@ -137,3 +140,30 @@ func TestCreateAttestationNoteAndOccurrence(t *testing.T) {
...
+
+	// Set PageSize to 300
+	createListOccurrencesRequest = createListOccurrencesRequest_test
+
+	vuln, err := d.Vulnerabilities("gcr.io/kritis-int-test/java-with-vulnz@sha256:358687cfd3ec8e1dfeb2bf51b5110e4e16f6df71f64fba01986f720b2fcba68a")
+	if err != nil {
+		t.Fatalf("Found err %s", err)
+	}
+
+	if len(vuln) <= 300 {
```
aysylu: With paging APIs, you aren't actually guaranteed to receive the full page of results, up to the max number you set. Therefore, you can't rely on this as a signal for whether there are more results. Instead, you'd want to check whether the next page token is set, to figure out if more pages of results are available.

pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
acamadeo: According to http://gcr.io/kritis-int-test/java-with-vulnz@sha256:358687cfd3ec8e1dfeb2bf51b5110e4e16f6df71f64fba01986f720b2fcba68a, there are 931 vulnerabilities as of now. I checked and the length of the vuln array is 931

pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
```diff
@@ -137,3 +140,30 @@ func TestCreateAttestationNoteAndOccurrence(t *testing.T) {
...
+	// Set PageSize to 300
+	createListOccurrencesRequest = createListOccurrencesRequestTest
+
+	vuln, err := d.Vulnerabilities("gcr.io/kritis-int-test/java-with-vulnz@sha256:358687cfd3ec8e1dfeb2bf51b5110e4e16f6df71f64fba01986f720b2fcba68a")
+	if err != nil {
+		t.Fatalf("Found err %s", err)
+	}
+
+	if len(vuln) <= 300 {
+		t.Errorf("Pagination error: only received results from first page.")
```
ooq: t.Fatalf?

pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
ooq: I don't actually fully understand what @aysylu meant here.

Apart from that, I think a more rigorous test is to check if len(vulnz) > 600, so we know for sure it goes over at least two pages.

ooq:
Thanks for the PR @acamadeo . I left some comments.

pkg/kritis/metadata/containeranalysis/containeranalysis.go
```diff
@@ -107,23 +120,19 @@ func (c Client) fetchVulnerabilityOccurrence(containerImage string, kind string)
...
 	}
 
-	req := &grafeas.ListOccurrencesRequest{
-		Filter:   fmt.Sprintf("resourceUrl=%q AND kind=%q", util.GetResourceURL(containerImage), kind),
-		PageSize: constants.PageSize,
-		Parent:   fmt.Sprintf("projects/%s", getProjectFromContainerImage(containerImage)),
-	}
+	req := createListOccurrencesRequest(containerImage, kind)
 
 	it := c.client.ListOccurrences(c.ctx, req)
```
ooq: Did the test actually fail with the old code? I can see that the grafeas client implements some pagination logic, so it would be great to verify that this patch is necessary.

pkg/kritis/metadata/containeranalysis/containeranalysis.go
acamadeo: According to the documentation on iterators (https://github.com/googleapis/google-cloud-go/wiki/Iterator-Guidelines#examples), iteration through pages requires you to use an `iterator.NewPager`. When you call its `NextPage` method, it adds the items from that page to a provided slice. It returns the `nextPageToken`, which will be empty once you have iterated through all the pages.

pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
acamadeo: In response to @aysylu,
I think my reasoning for this assertion is that `vuln` should contain all of the Vulnerabilities, and if the length of `vuln` exceeds the page size, then we must have extracted results from multiple pages. So it must have implicitly set the next page token.

I can try more directly testing whether the next page token is set, if you think that would be a better approach

acamadeo: In response to @ooq,
For a stronger assertion, I modified the page size to 100 and checked if len(vulnz) > 900. The page size is still large enough to not time out.

pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
acamadeo: I'm curious why t.Fatalf would be better here? My understanding was that the difference between t.Fatalf and t.Errorf is that the latter ends execution immediately. Would that matter here, as it is the last statement in the test?

pkg/kritis/metadata/containeranalysis/containeranalysis.go
acamadeo: This test still passes with the old code. There may be some benefit in using the new code, as it makes it more explicit that the pages are being "flipped" and that you have received all the results when the `nextPageToken` is blank.

pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
acamadeo: Done.
Should I apply the same change to the top-level test `TestGetMultiplePages_Vulnerabilities`?

pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
acamadeo: Done.

pkg/kritis/metadata/containeranalysis/containeranalysis.go
acamadeo: Reverted.

pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
ooq: Yes, logically here it's same, but just as a consistency to the rest of the code. And we never know if the difference could play out elsewhere.

pkg/kritis/metadata/containeranalysis/containeranalysis.go
ooq: I think the page flipping logic is already implemented in https://github.com/grafeas/kritis/blob/588f70e090ad2ee4c98e65a6d105976ade4aabb2/vendor/cloud.google.com/go/containeranalysis/apiv1beta1/grafeas_v1_beta1_client.go#L181
So it's actually desirable that the details is abstracted away (hidden from this function here).

pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
```diff
@@ -155,15 +155,15 @@ func TestGetMultiplePages_Vulnerabilities(t *testing.T) {
 		t.Fatalf("Found err %s", err)
 	}
 
-	if len(vuln) <= 300 {
-		t.Errorf("Pagination error: only received results from first page.")
+	if len(vuln) <= 900 {
+		t.Errorf("Pagination error: did not receive results from the final page.")
```
ooq: The error message can become incorrect when the vulnz count goes above 1000.

pkg/kritis/metadata/containeranalysis/containeranalysis.go
acamadeo: Done.

pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
acamadeo: Done.

pkg/kritis/metadata/containeranalysis/containeranalysis.go
acamadeo: Reverted to original implementation. Integration test shows that there is no bug while extracting vulnerabilities.

pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
acamadeo: Done.

ooq:
Thanks for the changes. I approved it. @aysylu Can you take another look? Thanks!

pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
aysylu: Re API for pagination, best practices: https://aip.dev/158 "The API may return fewer results than the number requested (including zero results), even if not at the end of the collection."

pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
```diff
@@ -137,3 +140,30 @@ func TestCreateAttestationNoteAndOccurrence(t *testing.T) {
...
 	}
 }
+
+func TestGetMultiplePagesVulnerabilities(t *testing.T) {
+	client, err := New()
+	if err != nil {
+		t.Fatalf("Could not initialize the client %s", err)
+	}
+
+	// Set PageSize to 300
```
ooq: mismatch.

acamadeo: Done.

pkg/kritis/metadata/containeranalysis/containeranalysis_int_test.go
acamadeo: I see, so I'd like to test that the next page token is being set, but none of our code actually does that. That's happening within the Container Analysis API library: https://github.com/grafeas/kritis/blob/588f70e090ad2ee4c98e65a6d105976ade4aabb2/vendor/cloud.google.com/go/containeranalysis/apiv1beta1/grafeas_v1_beta1_client.go#L188 

Our code enters an infinite loop until the client library to returns iterator.Done: https://github.com/grafeas/kritis/blob/588f70e090ad2ee4c98e65a6d105976ade4aabb2/pkg/kritis/metadata/containeranalysis/containeranalysis.go#L119.

Should I revise the test to make sure that we are receiving the iterator.Done signal?

PR #477, opened 2020-03-02
Remove public key in signer interface

ooq:
cc @donmccasland


donmccasland:
Okay, this makes a LOT more sense to me.  It didn't make sense that signing needed a public key.

Does this also fix the problem where we could only use an empty passphrase?

ooq:
Just added in https://github.com/grafeas/kritis/pull/478

PR #482, opened 2020-03-15
Fix may cause panic when getting metadata client failed

mozillazg:


aysylu:
@mozillazg Thanks for the fix, good catch. To clarify: why did you change the title of the PR to "may cause panic when ..." -- from what I can tell, this fixes the issue when getting metadata client fails, which is why I merged it. Is there something else I'm missing, that introduces a new failure mode?

cc @wkozlik

mozillazg:
@aysylu There are two ways to implement a `client.Close()` which do nothing:

* `(c Client) Close(){}`: panic
* `(c *Client) Close(){}`: not panic

PR #484, opened 2020-03-17
Don't logging pod struct which may contains sensitive data

mozillazg:
For example, the env of container may contains sensitive data in some cases.


aysylu:
@ooq and @wkozlik: could you PTAL and review?

pkg/kritis/admission/admission.go
```diff
@@ -297,7 +297,7 @@ func reviewImageSecurityPolicy(images []string, ns string, pod *v1.Pod, ar *v1be
 
 	r := admissionConfig.reviewer()
 	if err := r.ReviewISP(images, isps, pod, client); err != nil {
-		glog.Infof("Denying %s in namespace %s: %v", pod, ns, err)
+		glog.Infof("Denying %s in namespace %s: %v", getPodSummary(pod), ns, err)
```
wkozlik: Why not just log a `pod.Name`?
There are a couple instances in this file that do it that way.

mozillazg: ok, I'll change to use getPodName instead. why not just `pod.Name`, because the pod may be `nil` when review deployment or replica set.

PR #487, opened 2020-03-19
Merge RawAttestation Wrapper PR into master

acamadeo:
Merge changes from [Multiple key types - Create RawAttestation wrapper PR](https://github.com/grafeas/kritis/pull/474) into master


ooq:
Hi @acamadeo , thanks for the PR.
I suspect the only test that is failing to be flaky. Can you run the local e2e tests to verify? (or push some changes to this PR to trigger a re-test.)

PR #489, opened 2020-03-26
Refactor validating_transport_test and container_test 

acamadeo:
`validating_transport_test` contained one massive `TestValidatingTransport` test, which was testing multiple types of validation behavior, including proper key encoding, verification of signatures using keys, and the validated attestations returned by the transport.

Now, `TestValidatingTransport` is replaced by two test functions that test the public methods of the transport (i.e., `ValidatePublicKey` and `GetValidatedAttestation`). Additionally, signature verification testcases have been moved to container_test's `TestVerifySignature` method.


pkg/kritis/review/validating_transport.go
```diff
@@ -141,11 +142,11 @@ func (avt *AttestorValidatingTransport) GetValidatedAttestations(image string) (
 			continue
 		}
 		keyId := rawAtt.Signature.PublicKeyId
-		if err = host.VerifySignature(keys[keyId], string(decodedSig)); err != nil {
+		if err = avt.ContainerHost.VerifySignature(keys[keyId], string(decodedSig)); err != nil {
```
nenaddedic, line 145: OK, I think now I understand what you're trying to do.  I think you're trying to create an injection point here, so you can mock signature verification in unit tests.  Right?

I think that's a good idea, but requires a little different approach.  I think you'll end up having to define the crypto interface that we were discussing, and then mocking it up for tests.  I don't think ContainerHost is the thing you want to mock here.

PR #490, opened 2020-03-30
Fix cron integration test -- wait for pods to delete between testcases

acamadeo:
`TestKritisCron` was failing because it was not waiting for pods to terminate between testcases. Introduced a `kubectl wait` command between testcases.

### DIFF
--- a/integration/run_test.go
+++ b/integration/run_test.go
@@ -92,7 +92,7 @@ func cleanupTemplate(t *testing.T, path, ns string) error {
 		return fmt.Errorf("kubectl delete failed: %v\nout: %s", err, out)
 	}
 
-	cmd = exec.Command("kubectl", "wait", "--for=delete")
+	cmd = exec.Command("kubectl", "wait", "--for=delete", "--timeout=3m")
 	if out, err := integration_util.RunCmdOut(cmd); err != nil {
 		t.Fatalf("kubectl wait failed: %v\nout: %s", err, out)
 	}
--- a/integration/run_test.go
+++ b/integration/run_test.go
@@ -84,13 +84,17 @@ func cleanupTemplate(t *testing.T, path, ns string) error {
 		return nil
 	}
 	t.Logf("Cleaning up after %s ...", path)
-	cmd := exec.Command("kubectl", "delete", "-f", path, "--namespace", ns)
-	output, err := integration_util.RunCmdOut(cmd)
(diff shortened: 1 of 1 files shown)


integration/run_test.go
```diff
@@ -84,13 +84,17 @@ func cleanupTemplate(t *testing.T, path, ns string) error {
...
+	out, err := integration_util.RunCmdOut(cmd)
 	if err != nil {
-		return fmt.Errorf("kubectl delete failed: %s %v", output, err)
+		return fmt.Errorf("kubectl delete failed: %s %v", out, err)
 	}
-	if err := os.Remove(path); err != nil {
-		return fmt.Errorf("remove failed: %v", err)
+	cmd = exec.Command("kubectl", "wait", "--for=delete")
+	if out, err := integration_util.RunCmdOut(cmd); err != nil {
+		t.Fatalf("testing error: %v\nout: %s", err, out)
```
ooq: Update the message to `kubectl wait failed:`?

ooq:
Thanks for the fix! This is super helpful in improving the dev flow.
Left a comment in error message, otherwise LGTM.

integration/run_test.go
acamadeo: Done.

ooq:
It looks like now the test can timeout on the wait? @acamadeo

acamadeo:
> It looks like now the test can timeout on the wait? @acamadeo

It looks like it -- I could provide a `timeout` parameter to `kubectl wait`, which will resort to the behavior before the fix. How does a 3 minute wait sound?

ooq:
> > It looks like now the test can timeout on the wait? @acamadeo
> 
> It looks like it -- I could provide a `timeout` parameter to `kubectl wait`, which will resort to the behavior before the fix. How does a 3 minute wait sound?

I think there are two ways to go with the new 15minute timeout failure here:
- we might want to investigate and understand why it takes > 15min (if not forever) for the pod to terminate, which is probably why the test is flaky in the first place. Having a 3min timeout on the wait could just bring us back to ground zero and not fix this flakiness at all.
- the alternative fix to this is to ignore terminating pods during the label check for the subsequent test

WDYT?

liuplgtm:
looks like kokoro complains because of the wait_for issue. 

let us wait for that issue to be resolved first.

PR #492, opened 2020-04-15
Crypto Library Interface

acamadeo:
This is an interface for the new crypto library. It defines high-level functions for verifying and creating attestations. These functions are accessible through separate Go interfaces: Verifier and Signer, respectively.


cryptolib/attestation.go
```diff
@@ -0,0 +1,89 @@
...
+	// ECDSA on the NIST P-384 curve with a SHA384 digest.
+	EcdsaP384Sha384
+	// ECDSA on the NIST P-521 curve with a SHA512 digest.
+	EcdsaP521Sha512
+)
+
+// Attestation is a generic wrapper for an attestation. It can store signatures
+// generated by PGP or PKIX keys. Alternatively, it can store an attestation
+// represented as a JWT.
+type Attestation struct {
```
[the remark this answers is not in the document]
nenaddedic: And, do you imagine some helper methods to construct Attestation objects from ATTESTATION Occurrences?

cryptolib/verifier.go
```diff
@@ -0,0 +1,59 @@
...
+type verifier struct{}
+
+// NewVerifier creates a Verifier interface.
+func NewVerifier() Verifier {
+	return &verifier{}
+}
+
+// VerifyAttestation verifies whether an Attestation satisfies an
+// AttestationAuthority's public key under an image.
+func (v *verifier) VerifyAttestation(att *Attestation, publicKey []byte, image string) error {
```
nenaddedic, line 35: Public key sometimes comes with an id (PGP), but sometimes doesn't (both PKIX and JWT, I think).  During verification, we will require that key IDs match, not just that the signature verifies.

So -- while key ID matching could happen outside of this function, at the calling site, I'd consider pulling it in here explicitly.  This would require adding a key id argument in some form.

It could be an explicit argument, and then the relationship between key id and publicKey should be clarified (e.g. for PGP).  You could also create a dedicated PublicKey interface which encapsulates both the id and the key material.

Potential added benefit of a dedicated PublicKey interface is that it further decouples VerifyAttestation from wire format.  But that's a bit unclear still until you document how publicKey argument should be set.

cryptolib/attestation.go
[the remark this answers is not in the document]
acamadeo: Yes, I was imagining this to replace Kritis's `RawAttestation`. It would be helpful to have methods that parse this struct from Attestation Occurrences, but I feel that it probably does not belong in this `cryptolib` package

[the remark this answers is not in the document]
acamadeo: Perhaps it makes sense to have an `attestation` package that would contain this struct and any helper methods in this crypto library. WDYT?

acamadeo:
In case it is helpful, here is snippet with an example of how we might use this interface:
https://pastebin.com/JXabVVeB

cryptolib/attestation.go
```diff
@@ -0,0 +1,89 @@
...
+// Enumeration of KeyMode
+const (
+	UnknownKeyMode KeyMode = iota
+	Pgp
+	Pkix
+	Jwt
+)
+
+// Enumeration of SigningAlgorithm
+const (
```
ooq, line 36: Are the signing algorithms only for PKIX? If so, should we move them to a PKIX file?

alexcope: +1

cryptolib/signer.go
```diff
@@ -0,0 +1,56 @@
...
+limitations under the License.
+*/
+
+package cryptolib
+
+import "errors"
+
+// Signer contains methods to create a signed Attestation.
+type Signer interface {
+	CreateAttestation(privateKey, payload []byte, publicKeyID string, alg SigningAlgorithm, keyMode KeyMode) (*Attestation, error)
```
alexcope, line 23: nit: privateKey []byte, ....

Explicitly specificity the type of each function argument unless the signatures are very simple (e.x function takes 4 strings and nothing else)

cryptolib/attestation.go
```diff
@@ -0,0 +1,89 @@
...
+	Signature Signature
+	// SerializedPayload stores the payload over which the signature was
+	// signed. This field is only used for Pkix Attestations. For Pgp and Jwt
+	// Attestations, the payload is bundled with the signature itself.
+	SerializedPayload []byte
+}
+
+// Signature holds the signature content of the Attestation.
+type Signature struct {
+	// LegacySignature stores the signature for Pgp and Pkix Attestations.
```
alexcope, line 85: I don't like this implicit oneof struct where different signatures have different types.

Two potential solutions

1. If possible, make all signatures either a byte array or a string
2. Otherwise, make this an interface and have two types that implement the interface.

cryptolib/signer.go
alexcope, line 23: Document all function in the interface. Individual implementations document things specific to the implementation and can reference the documentation in the interface. Ditto all interfaces in this PR

cryptolib/attestation.go
```diff
@@ -0,0 +1,89 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cryptolib
+
+type (
+	// KeyMode stores the type of the Attestation.
+	KeyMode int
```
alexcope, line 21: Key Mode is entirely determined by Signature Algorithm, I'd remove it and update Signature Algorithm to be pkix_*, pgp_* etc

cryptolib/attestation.go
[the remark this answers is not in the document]
ooq: I can see the parsing helpers to be "librarized", but probably not in scope for the crypto library now?
+ for having attestation struct in its own package though.

cryptolib/signer.go
ooq, line 23: I think `privateKey` should be held in a config-like struct in the signer, instead of passing it as a parameter. The same should also apply to other metadata that are likely to remain constant for a signer instance, e.g., `SigningAlgorithm` or `keyMode`, `publicKeyID`.

cryptolib/signer.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+type signer struct{}
+
+// CreateAttestation creates an Attestation whose signature is generated by
+// signing the given payload with the private key. For PGP and PKIX, `payload`
+// should be the raw payload data. For JWT, `payload` should be a serialized
+// but unsigned token.
+func (s *signer) CreateAttestation(privateKey, payload []byte, publicKeyID string, alg SigningAlgorithm, keyMode KeyMode) (*Attestation, error) {
```
ooq, line 32: Just same comment as above, depending on how `signer` is instantiated. 
An alternative to this is having `CreateAttestation` as a method in the interface, and separating implemented by each Signer type.
Not suggesting this is the way to do it, but just throwing out an alternative here.

cryptolib/attestation.go
treyridley: Anthony and I discussed this earlier today and  thought of KeyMode having a SignatureAlgorithm field. Would you be opposed to this rather than adding the KeyMode as a prefix in the SignatureAlgorithm?

cryptolib/attestation.go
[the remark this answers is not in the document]
acamadeo: I would agree, I don't think it is necessary at the moment. When we integrate the library into Kritis, we will likely need to do this. Especially if we are converting `Occurrences` into `Attestations`.

cryptolib/signer.go
acamadeo: This would make sense to me if a user would want to create multiple attestations using the same private key.

For example, I can imagine it would be useful if the signer tool will be used to sign containers "in bulk" (e.g. a vulnerability signer processing a queue of container images). But I am not sure that this matches our current/near-future use case. Since you are working on the signer, could you provide more context on this?

cryptolib/attestation.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Attestation is a generic wrapper for an attestation. It can store signatures
+// generated by PGP or PKIX keys. Alternatively, it can store an attestation
+// represented as a JWT.
+type Attestation struct {
+	// KeyMode is the type of the Attestation: one of Pgp, Pkix, or Jwt.
+	KeyMode KeyMode
+	// SigningAlgorithm is the public-key signing algorithm and hashing
+	// function used to sign the Attestation.
+	SigningAlgorithm SigningAlgorithm
```
patricklawsongoogle, line 72: We should be inferring this from the public key (trust anchor) configuration.  We don't trust the signing algorithm provided by the attestation, so there isn't really a point in keeping it around.  Also note that in order to fill this out properly when converting from a JWT, you'd need to parse the JWT header to find the value.  (The same is also true for PGP signatures, but more difficult to obtain and even less useful).

cryptolib/verifier.go
```diff
@@ -0,0 +1,59 @@
...
+limitations under the License.
+*/
+
+package cryptolib
+
+import "errors"
+
+// Verifier contains methods to validate an Attestation.
+type Verifier interface {
+	VerifyAttestation(att *Attestation, publicKey []byte, image string) error
```
patricklawsongoogle, line 23: I would prefer to see this split into two layers: 

1. Low level cryptographic authentication (i.e. `att`'s signature is validated by `publicKey`).
2. High level verification in the context of the current policy evaluation (i.e. `att` is both cryptographically authenticated *and* its payload correctly verifies `image` according to Kritis semantics).

(2) would call (1) as an implementation detail.

cryptolib/verifier.go
patricklawsongoogle: +1 to a dedicated PublicKey interface.  That will make it easy to enforce that any attestation-provided signing algorithm isn't accidentally trusted for the purpose of signature verification.  It should also make testing easier.

It's also easy to imagine a PublicKeySet with the same interface that encapsulates a bundle of PublicKey instances and returns iff at least one of the encapsulated PublicKeys verifies the attestation.  It could also be responsible for dispatching to the right PublicKey based on keyId, and rejecting if it doesn't know about any such keyId.

cryptolib/verifier.go
```diff
@@ -0,0 +1,59 @@
...
+func NewVerifier() Verifier {
+	return &verifier{}
+}
+
+// VerifyAttestation verifies whether an Attestation satisfies an
+// AttestationAuthority's public key under an image.
+func (v *verifier) VerifyAttestation(att *Attestation, publicKey []byte, image string) error {
+	switch att.KeyMode {
+	case Pkix:
+		return verifyPkixAndCheckPayload(att, image)
```
ooq, line 38: Missing public key?

Also just follow up to the comment above on PublicKey interface, I also think that once `keymode` is decided, the code should only work on the wire-format crypto level, e.g., it does not need to understand attestation, but only work on signatures, etc.

cryptolib/verifier.go
ooq: +1 to what's said above, 1) the crypto library should also include the logic of public key match-up by key id, and 2) once key is looked up, we only need to work on the crypto level, without worrying about high-level concepts like attestation.

aysylu:
@wkozlik just as an FYI

cryptolib/attestation.go
alexcope: I'm note sure I follow. Do you mean creating a type which contains the information in KeyMode and SigningAlgoritm? That sounds reasonable to me.

Also small nit: SigningAlgoritm-> SignatureAlgorithm becaues it is used in verification as well.

cryptolib/attestation.go
wkozlik: Wouldn't the `KeyMode` already determine if the signature is JWT or not? I am wondering if a single field `Signature []byte` would suffice.

cryptolib/signer.go
```diff
@@ -0,0 +1,70 @@
...
+	case Jwt:
+		return createJwtAttestation(payload, alg)
+	default:
+		return nil, errors.New("invalid key mode")
+	}
+}
+
+// Unimplemented functions
+func createPkixAttestation(payload []byte, alg SignatureAlgorithm) (*Attestation, error) {
+	return nil, errors.New("attestation type not implemented")
```
ooq, line 61: nit: attestation type Pkix not implemented.

cryptolib/signer.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func createPgpAttestation(payload []byte, alg SignatureAlgorithm) (*Attestation, error) {
+	return nil, errors.New("attestation type not implemented")
```
ooq, line 65: nit: attestation type Pgp not implemented.

cryptolib/signer.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func createJwtAttestation(payload []byte, alg SignatureAlgorithm) (*Attestation, error) {
+	return nil, errors.New("attestation type not implemented")
```
ooq, line 69: nit: attestation type Jwt not implemented.

cryptolib/verifier.go
acamadeo: We have encapsulated the key data and id in a `PublicKey` struct. We also modified our `VerifyAttestation` function to receive `publicKeySet`, a list of `PublicKeys`.

cryptolib/attestation.go
acamadeo: We opted to put all the algorithms together in a single `signature_algorithm.go` file. This should make it easy to add/remove new algorithms without digging into other files to adjust enum values.

cryptolib/signer.go
acamadeo: Done.

cryptolib/attestation.go
acamadeo: Done. The field is now `Signature []byte`.

cryptolib/verifier.go
```diff
@@ -0,0 +1,103 @@
...
+	case Jwt:
+		payload, err = verifyJwt(att.Signature, publicKey.KeyData)
+	default:
+		return errors.New("signature uses an unsupported key mode")
+	}
+	if err != nil {
+		return err
+	}
+
+	expected := Metadata{image}
```
ooq, line 73: What does this do?

cryptolib/signer.go
acamadeo: Done.

cryptolib/attestation.go
acamadeo: Done. We opted for the prefix solution that you suggested. We also changed the name to `SignatureAlgorithm`.

cryptolib/attestation.go
acamadeo: Done. We removed the `KeyMode` and `SignatureAlgorithm` fields.

cryptolib/verifier.go
acamadeo: Done. We added `verifyPkix`, `verifyPgp`, and `verifyJwt` functions to explicitly highlight the low level crypto layer.

cryptolib/verifier.go
acamadeo: Done. Our low level crypto (e.g. `verifyPkix`, etc.) no longer receives an attestation -- it now receives the minimum information needed to verify the signature.

cryptolib/signer.go
```diff
@@ -0,0 +1,70 @@
...
+
+type signer struct {
+	PrivateKey  []byte
+	PublicKeyID string
+}
+
+// NewSigner creates a Signer interface. `privateKey` contains the key data for
+// the private key that signs attestations. `publicKeyID` is the ID of the
+// public key that can verify those signatures.
+func NewSigner(privateKey []byte, publicKeyID string) Signer {
```
nenaddedic, line 38: Will this function be able to decide if it should create a PGP, PKIX, or JWT signer?  It seems like it's missing a hint about this.

Also the format of privateKey should be clearly documented (it's OK to do this as we add implementations, but let's just make sure we don't forget).  For example, in case of PGP, is it
(i) ascii armored, or
(ii) ascii armored and then base64 encoded, or
(iii) something else?

cryptolib/verifier.go
```diff
@@ -0,0 +1,103 @@
...
+	// KeyData holds the raw key material which can verify a signature.
+	KeyData []byte
+	// ID uniquely identifies this public key. For PGP, this should be the
+	// OpenPGP RFC4880 V4 fingerprint of the key.
+	ID string
+}
+
+type verifier struct{}
+
+// NewVerifier creates a Verifier interface.
```
nenaddedic, line 41: Nit: creates a Verifier instance
(right?)

cryptolib/signer.go
```diff
@@ -20,37 +20,51 @@ import "errors"
...
+}
 
-// CreateAttestation creates an Attestation whose signature is generated by
-// signing the given payload with the private key. For PGP and PKIX, `payload`
-// should be the raw payload data. For JWT, `payload` should be a serialized
-// but unsigned token.
-func (s *signer) CreateAttestation(privateKey, payload []byte, publicKeyID string, alg SigningAlgorithm, keyMode KeyMode) (*Attestation, error) {
-	switch keyMode {
+// CreateAttestation creates a signed Attestation. See Signer for more details.
+func (s *signer) CreateAttestation(payload []byte, alg SignatureAlgorithm) (*Attestation, error) {
```
alexcope, line 46: Nit: How common is it to use multiple algorithms with a single key? If that is not common I think it makes sense to move the alg field into NewSigner.

cryptolib/signature_algorithm.go
```diff
@@ -0,0 +1,111 @@
...
+	// PGP Signing Algorithms
+	//
+	// RSASSA-PKCS1-v1_5 with a 2048 bit key and a SHA256 digest.
+	PgpRsaSignPkcs12048Sha256
+	// RSASSA-PKCS1-v1_5 with a 3072 bit key and a SHA256 digest.
+	PgpRsaSignPkcs13072Sha256
+	// RSASSA-PKCS1-v1_5 with a 4096 bit key and a SHA256 digest.
+	PgpRsaSignPkcs14096Sha256
+	// RSASSA-PKCS1-v1_5 with a 4096 bit key and a SHA512 digest.
+	PgpRsaSignPkcs14096Sha512
```
patricklawsongoogle, line 62: PGP's signature algorithms are encapsulated within the public key itself; we shouldn't need these.

cryptolib/signature_algorithm.go
```diff
@@ -0,0 +1,111 @@
...
+	JwtRsaSignPkcs14096Sha256
+	// RSASSA-PKCS1-v1_5 with a 4096 bit key and a SHA512 digest.
+	JwtRsaSignPkcs14096Sha512
+
+	// ECDSA on the NIST P-256 curve with a SHA256 digest.
+	JwtEcdsaP256Sha256
+	// ECDSA on the NIST P-384 curve with a SHA384 digest.
+	JwtEcdsaP384Sha384
+	// ECDSA on the NIST P-521 curve with a SHA512 digest.
+	JwtEcdsaP521Sha512
```
patricklawsongoogle, line 89: We shouldn't need to distinguish between JWT and PKIX here.  All of these signature algorithms are specifically for PKIX, and we just happen to support all of the PKIX signature algorithms for JWT under the hood.

cryptolib/signer.go
```diff
@@ -0,0 +1,70 @@
...
+	case Pkix:
+		return createPkixAttestation(payload, alg)
+	case Pgp:
+		return createPgpAttestation(payload, alg)
+	case Jwt:
+		return createJwtAttestation(payload, alg)
+	default:
+		return nil, errors.New("invalid key mode")
+	}
+}
```
patricklawsongoogle, line 57: I'm not sure that this is worth the trouble versus just letting the caller call the CreateFooAttestation functions directly.

cryptolib/signing_algorithm.go
```diff
@@ -0,0 +1,111 @@
...
+const (
+	UnknownKeyMode keyMode = iota
+	Pgp
+	Pkix
+	Jwt
+)
+
+func (sa SignatureAlgorithm) keyMode() keyMode {
+	switch {
+	case sa >= PkixRsaPss2048Sha256 && sa < PgpRsaSignPkcs14096Sha512:
```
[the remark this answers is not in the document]
ooq: Not sure why this comment becomes outdated ... bumping it.

cryptolib/verifier.go
```diff
@@ -0,0 +1,103 @@
...
+
+import "errors"
+
+// Verifier contains methods to validate an Attestation.
+type Verifier interface {
+	// VerifyAttestation verifies whether an Attestation satisfies at least one
+	// of the public keys under an image. This function finds the public key
+	// whose ID matches the attestation's PublicKeyID, and uses this key to
+	// verify the signature.
+	VerifyAttestation(att *Attestation, publicKeySet []PublicKey, image string) error
```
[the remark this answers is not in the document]
ooq: Same here. Bumping.

cryptolib/signer.go
```diff
@@ -0,0 +1,70 @@
...
+// Signer contains methods to create a signed Attestation.
+type Signer interface {
+	// CreateAttestation creates an Attestation whose signature is generated by
+	// signing the given payload with the private key. For PGP and PKIX, `payload`
+	// should be the raw payload data. For JWT, `payload` should be a serialized
+	// but unsigned token.
+	CreateAttestation(payload []byte, alg SignatureAlgorithm) (*Attestation, error)
+}
+
+type signer struct {
```
ooq, line 30: A question is whether we want a signer instance to take the key type or algorithm.

cryptolib/signer.go
```diff
@@ -0,0 +1,70 @@
...
+// public key that can verify those signatures.
+func NewSigner(privateKey []byte, publicKeyID string) Signer {
+	return &signer{
+		PrivateKey:  privateKey,
+		PublicKeyID: publicKeyID,
+	}
+}
+
+// CreateAttestation creates a signed Attestation. See Signer for more details.
+func (s *signer) CreateAttestation(payload []byte, alg SignatureAlgorithm) (*Attestation, error) {
```
ooq, line 46: Since `CreateAttestation` is a general interface, one can image that each type-specific signer would overwrite this method? In that case, we don't need this function for key-based branching.

Alternatively, if we don't need type-specific signers, it does make sense to just let the caller to call CreateAttestation directly, providing all necessary parameters (type, algorithm, privatekey, image) to do so.

cryptolib/signer.go
```diff
@@ -0,0 +1,92 @@
...
+		PrivateKey:         privateKey,
+		PublicKeyID:        publicKeyID,
+		SignatureAlgorithm: alg,
+	}
+}
+
+// NewPgpSigner creates a Signer interface for PGP Attestations. `privateKey`
+// contains the ASCII-armored private key. `publicKeyID` is the ID of the
+// public key that can verify the Attestation signature.
+func NewPgpSigner(privateKey []byte, publicKeyID string) Signer {
```
nenaddedic, line 61: Is publicKeyID derivable from privateKey?

cryptolib/signer.go
```diff
@@ -0,0 +1,92 @@
...
+type jwtSigner struct {
+	PrivateKey         []byte
+	PublicKeyID        string
+	SignatureAlgorithm SignatureAlgorithm
+}
+
+// NewPkixSigner creates a Signer interface for PKIX Attestations. `privateKey`
+// contains the PEM-encoded private key. `publicKeyID` is the ID of the public
+// key that can verify the Attestation signature.
+func NewPkixSigner(privateKey []byte, publicKeyID string, alg SignatureAlgorithm) Signer {
```
nenaddedic, line 50: This, and other factory methods, should return (Signer, error).  So if the key cannot be parsed or something you can return an error.

cryptolib/verifier.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// PublicKey stores public key material for all key types.
+type PublicKey struct {
```
nenaddedic, line 31: How do I create a PublicKey?  Signer has NewThisSinger and NewThatSigner, which makes it nice and obvious how to create one, but there's no such obvious counterpart here.

cryptolib/verifier.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// AuthenticatedAttestation contains data that is extracted from an Attestation
+// only after its signature has been verified. The contents of an Attestation
+// payload should never be analyzed directly, as it may or may not be verified.
+// Instead, these should be extracted into an AuthenticatedAttestation and
+// analyzed from there.
+// NOTE: The concept and usefulness of an AuthenticatedAttestation are still
+// under discussion and is subject to change.
+type AuthenticatedAttestation struct {
```
nenaddedic, line 98: Seems like this  is not intended to be exported, no?

cryptolib/signer.go
alexcope: In a sane world yes, but binauthz allows users to assign arbitrary ID's to their keys

cryptolib/signer.go
acamadeo: Done.

cryptolib/signer.go
acamadeo: Done.

cryptolib/signer.go
acamadeo: Done.

cryptolib/signing_algorithm.go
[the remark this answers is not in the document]
acamadeo: Removed. This conversion function is no longer necessary.

cryptolib/verifier.go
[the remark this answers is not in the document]
acamadeo: Name changed. It receives an image digest.

cryptolib/verifier.go
acamadeo: We replaced `Metadata` with `AuthenticatedAttestation`. In essence, `AuthenticatedAttestation` stores metadata from the payload of an attesatation that has already been verified. I added a docstring above the struct definition with more details.

cryptolib/signer.go
acamadeo: We have opted to create separate constructors for each of the key modes. We also added descriptions for the formatting of the private key in the PGP and PKIX case. We selected the same formatting that Binauthz expects of the corresponding public key. We will provide documentation for JWT keys as we near/complete implementation.

cryptolib/verifier.go
acamadeo: I modified the docstring to explain what the interface is for.

cryptolib/signer.go
acamadeo: Done. To be clear, each key mode's signer is now implemented by its own type. We added the SignatureAlgorithm parameter to the PKIX and JWT constructors. We did not include it in the PGP constructor, as the algorithm can be inferred directly from the private key.

cryptolib/signature_algorithm.go
acamadeo: Done.

cryptolib/signature_algorithm.go
acamadeo: Done.

cryptolib/signer.go
acamadeo: We created separate signer implementations for each key type. See https://github.com/grafeas/kritis/pull/492#discussion_r414135858

cryptolib/signer.go
acamadeo: We created separate signer implementations for each key type. See https://github.com/grafeas/kritis/pull/492#discussion_r414135858

cryptolib/attestation.go
```diff
@@ -0,0 +1,49 @@
...
+	UnknownKeyMode keyMode = iota
+	Pgp
+	Pkix
+	Jwt
+)
+
+// TODO: Provide an implementation that determines the key mode from the
+// Attestation's signature.
+func (att *Attestation) extractKeyMode() keyMode {
+	return UnknownKeyMode
```
alexcope, line 48: I don't think this is possible for raw PKIX attestations.

cryptolib/signer.go
patricklawsongoogle: That's true on the enforcement side of things, but we can be more opinionated here if we wish.  I think it would be fine to start out by saying that the PGP signer always derives the key ID from the private key.  (We can always loosen that up later, but it's harder to go in the other direction)

acamadeo: Done.

cryptolib/verifier.go
acamadeo: Done.

cryptolib/verifier.go
acamadeo: Done.

cryptolib/attestation.go
acamadeo: Changed this so it extracts the `Pkix` key mode from the structure of the `Attestation`. That is, it will assume the `Attestation` is PKIX if `SerializedPayload` is nonempty, as only PKIX uses that field.

cryptolib/signer.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+type pkixSigner struct {
+	PrivateKey         []byte
+	PublicKeyID        string
+	SignatureAlgorithm SignatureAlgorithm
+}
+
+type pgpSigner struct {
+	PrivateKey []byte
```
ooq, line 37: We will still need the PublicKeyId for pgpSigner, so that the created attestation will contain it.

cryptolib/verifier.go
nenaddedic: OK, now I see what you're doing.  PublicKey doesn't "know" which keyMode it is.  Instead, when you verify an attestation, you deduce the keyMode from the attestation, and then try to parse the key.  This is different than the signer, where key mode is fixed at signer creation time (rather than later at signing time).

I wonder why the asymmetry?  It seems nicer to parse the key once, as early as possible.  It gives you some optimization opportunities, and also allows earlier error signaling.

Also -- supposing we ever add even more key types, will it always be possible to determine key type from attestation?

This is a non-blocking comment, if you want to stick to the current design, I don't feel strongly about it.  I just wonder if you'd thought about the alternative.

cryptolib/verifier.go
```diff
@@ -0,0 +1,122 @@
...
+type PublicKey struct {
+	// KeyData holds the raw key material which can verify a signature.
+	KeyData []byte
+	// ID uniquely identifies this public key. For PGP, this should be the
+	// OpenPGP RFC4880 V4 fingerprint of the key.
+	ID string
+}
+
+// NewPublicKey creates a new PublicKey.
+func NewPublicKey(keyData []byte, keyID string) PublicKey {
```
nenaddedic, line 40: If you leave this design, then please document the arguments.  How to fill them in, for each case:
- pgp
- pkix
- jwt

ooq:
Thanks for the PR and the updates! @acamadeo @treyridley 
Approving this. Let's decide on the code destination and identifying the kokoro failure before merging.

cryptolib/verifier.go
```diff
@@ -0,0 +1,122 @@
...
+
+type verifier struct{}
+
+// NewVerifier creates a Verifier interface for verifying Attestations.
+func NewVerifier() Verifier {
+	return &verifier{}
+}
+
+// VerifyAttestation verifies an Attestation. See Verifier for more details.
+func (v *verifier) VerifyAttestation(att *Attestation, publicKeySet []PublicKey, imageDigest string) error {
```
alexcope, line 55: As discussed in standup, make publicKeySet part of verifier. 
Also clarify imageDigest is trusted information and should not be just pulled off the attestation.

cryptolib/signer.go
acamadeo: Done.

cryptolib/signer.go
acamadeo: Done.

cryptolib/verifier.go
acamadeo: Done.

cryptolib/verifier.go
acamadeo: Done.

pkg/kritis/cryptolib/verifier.go
```diff
@@ -0,0 +1,148 @@
...
+// Verifier contains methods to validate an Attestation.
+type Verifier interface {
+	// VerifyAttestation verifies whether an Attestation satisfies at least one
+	// of the public keys under an image. This function finds the public key
+	// whose ID matches the attestation's PublicKeyID, and uses this key to
+	// verify the signature.
+	VerifyAttestation(att *Attestation) error
+}
+
+type KeyType int
```
nenaddedic, line 30: If signature algorithm has a dedicated file, then should this also have it?  Same file maybe?

PR #496, opened 2020-04-24
Add PGP Verification to Cryptolib

acamadeo:
Implement the `verifyPgp` function, which verifies an attached PGP signature using a public key and returns the payload that was signed.


cryptolib/pgp_test.go
```diff
@@ -0,0 +1,93 @@
...
+		{
+			name:           "invalid signature",
+			signature:      []byte("invalid-sig"),
+			publicKey:      readFile(t, "publicKey.gpg"),
+			payload:        readFile(t, "payload.txt"),
+			validSignature: false,
+			validPayload:   false,
+		},
+		{
+			name:           "invalid payload",
```
ooq, line 57: Is this test useful? The tested logic seems to be in the testing file.

acamadeo: Which file are you referring to? 

I believe this test is useful because we want to make sure that `verifyPgp` correctly extracts the payload from the signature. In this case, `signature.gpg` was created by signing over `payload.txt`, so we want to make sure the returned contents of `verifyPgp` match the contents of `payload.txt`.

acamadeo: Done. Removed

cryptolib/pgp_test.go
```diff
@@ -0,0 +1,93 @@
...
+			signature:      readFile(t, "signature.gpg"),
+			publicKey:      readFile(t, "publicKey.gpg"),
+			payload:        []byte("invalid-payload"),
+			validSignature: true,
+			validPayload:   false,
+		},
```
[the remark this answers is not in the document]
acamadeo: Done.

cryptolib/pgp_test.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	for _, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			payload, err := verifyPgp(tc.signature, tc.publicKey)
+			testutil.CheckError(t, !tc.validSignature, err)
```
alexcope, line 61: The if err != nil { return } is not something I see often in test code. I'd just use if-else to handle the expected error and expected success cases seperately.

alexcope, line 61: tc.validSignature  -> tc.expectedErr

cryptolib/pgp_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+			// Skip payload check if signature verification failed
+			if err != nil {
+				return
+			}
+
+			if tc.validPayload && string(payload) != string(tc.payload) {
+				t.Fatalf("Expected valid payload: got: %s, want: %s", string(payload), string(tc.payload))
```
alexcope: t.Errorf. Only use fatal if the rest of your test can not function if the condition is not met.

cryptolib/pgp.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/pkg/errors"
+	"golang.org/x/crypto/openpgp"
+	"golang.org/x/crypto/openpgp/armor"
+)
+
+// verifyPgp verifies a PGP signature using a public key and outputs the
+// payload that was signed. `signature` is an ASCII-armored "attached"
+// signature, generated by `gpg --armor --sign --output signature payload`.
+// `publicKey` is an ASCII-armored PGP key.
+func verifyPgp(signature []byte, publicKey []byte) ([]byte, error) {
```
alexcope, line 32: nit: signature, publicKey []byte

cryptolib/testdata/publicKey.gpg
```diff
@@ -0,0 +1,41 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
alexcope, line 1: nit: Make these constants in the test file instead of seperate test data

cryptolib/pgp_test.go
acamadeo: Done.

acamadeo: Done.

cryptolib/pgp_test.go
acamadeo: Done.

cryptolib/pgp.go
acamadeo: Done.

cryptolib/testdata/publicKey.gpg
acamadeo: Done.

PR #498, opened 2020-04-28
Fix issues in signer demo

ooq:
Fix several issues in signer demo:
- enable container scanning so vulnz results can be generated
- image should use `RepoDigests` instead of `ID` field in docker metadata. (Image ID is not digest)
- other minor fixes to make it work


ooq:
Hi @donmccasland, can you take a look of the fixes to see if they make sense?

ooq:
There is still issue with signer demo. The created attestation is base64 encoded and thus does not work naively with binary authorization.
As documented in #506.
Will fix this in a separate PR.

Merging this once test is passed.

PR #499, opened 2020-05-01
Add key lookup to crypto interface

acamadeo:
`VerifyAttestation` attempts to verify an attestation using the key ID specified in the Attestation.

Issue: #501


pkg/kritis/cryptolib/verifier.go
```diff
@@ -70,23 +81,32 @@ func NewVerifier(imageDigest string, publicKeySet []PublicKey) (Verifier, error)
...
+		payload   []byte
+		publicKey PublicKey
 	)
 
 	// Extract the public key from `publicKeySet` whose ID matches the one in
 	// `att`.
-	// TODO: Replace no-op with correct implementation.
-	publicKey := v.PublicKeySet[0]
+	foundKey := false
+	for _, key := range v.PublicKeySet {
```
nenaddedic, line 92: Consider changing the verifier struct so that PublicKeySet becomes a map:
id -> PublicKey
(And you'd have to change NewVerifier() to populate this map.)

If you do this, then you can avoid this loop and just do a map lookup.

pkg/kritis/cryptolib/verifier.go
```diff
@@ -70,23 +81,32 @@ func NewVerifier(imageDigest string, publicKeySet []PublicKey) (Verifier, error)
...
+		if key.ID == att.PublicKeyID {
+			publicKey, foundKey = key, true
+			break
+		}
+	}
+	if !foundKey {
+		return fmt.Errorf("Verifier doesn't contain matching public key with ID %s", att.PublicKeyID)
```
nenaddedic, line 99: This will probably be a user visible error message, so make sure its text doesn't contain anything which requires understanding Kritis internals ("Verifier" for example is an internal term).  Maybe just: "no public key with ID %s found" or something,

pkg/kritis/cryptolib/verifier_test.go
```diff
@@ -0,0 +1,75 @@
...
+			expectedErr: false,
+		},
+		{
+			// This is a possibility with user-provided IDs
+			name:        "different keys with same ID",
+			att:         att,
+			publicKeys:  []PublicKey{matchingKey, otherMatchingKey},
+			expectedErr: false,
+		},
+		{
```
nenaddedic, line 55: add another case with one matching and one non-matching key

pkg/kritis/cryptolib/verifier.go
```diff
@@ -16,7 +16,18 @@ limitations under the License.
...
-import "errors"
+import (
+	"errors"
+	"fmt"
+)
+
+// TODO: Remove function stubs once implemented
+// For testing
+var (
+	pkixVerify                      = verifyPkix
```
alexcope, line 27: I think you are trying to alias functions so you can stub them out for testing. Instead of using global variables for this, have verifier contain a private interface(s) which it uses. In the non-test case, when you construct a verifier it should automatically set the contained interface to the real implementation, but your unit tests can reach into the internals and replace it with a stub.

pkg/kritis/cryptolib/verifier.go
```diff
@@ -16,7 +16,18 @@ limitations under the License.
 
 package cryptolib
 
-import "errors"
+import (
+	"errors"
+	"fmt"
+)
+
+// TODO: Remove function stubs once implemented
```
alexcope, line 24: Always have a name or bug attached to TODOs

pkg/kritis/cryptolib/verifier.go
```diff
@@ -70,23 +81,32 @@ func NewVerifier(imageDigest string, publicKeySet []PublicKey) (Verifier, error)
 // VerifyAttestation verifies an Attestation. See Verifier for more details.
 func (v *verifier) VerifyAttestation(att *Attestation) error {
 	var (
-		err     error
-		payload []byte
+		err       error
```
alexcope, line 84: In go initialize local variables at time of first assignment with :=. the only common exception is if they are needed outside the scope where they are first assigned, in which case you should declare them as close to first assignment as possible.

pkg/kritis/cryptolib/verifier.go
ooq: Would you suggest using github issue as the bug system?

pkg/kritis/cryptolib/verifier.go
```diff
@@ -70,23 +81,32 @@ func NewVerifier(imageDigest string, publicKeySet []PublicKey) (Verifier, error)
...
+		payload   []byte
+		publicKey PublicKey
 	)
 
 	// Extract the public key from `publicKeySet` whose ID matches the one in
 	// `att`.
-	// TODO: Replace no-op with correct implementation.
-	publicKey := v.PublicKeySet[0]
+	foundKey := false
```
ooq, line 91: I think you can avoid this variable by checking `publicKey == nil`

pkg/kritis/cryptolib/verifier.go
```diff
@@ -70,23 +81,32 @@ func NewVerifier(imageDigest string, publicKeySet []PublicKey) (Verifier, error)
...
+		return fmt.Errorf("Verifier doesn't contain matching public key with ID %s", att.PublicKeyID)
+	}
 
 	switch publicKey.KeyType {
 	case Pkix:
-		err = verifyPkix(att.Signature, att.SerializedPayload, publicKey.KeyData)
+		err = pkixVerify(att.Signature, att.SerializedPayload, publicKey.KeyData)
 		payload = att.SerializedPayload
 	case Pgp:
 		payload, err = verifyPgp(att.Signature, publicKey.KeyData)
```
ooq, line 107: Just curious why is pgp not stubbed?

pkg/kritis/cryptolib/verifier.go
alexcope: That would work. I'd probably just use my ldap here because it's a little easier.

ooq: ah, you mean assignee/reporter's name, that totally makes sense!

pkg/kritis/cryptolib/verifier_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"testing"
+)
+
+func TestVerifyAttestation(t *testing.T) {
+	att := &Attestation{
+		PublicKeyID:       "key-id",
+		Signature:         []byte("signature"),
+		SerializedPayload: []byte("payload"),
+	}
+	matchingKey := NewPublicKey(Pkix, []byte("key-data"), "key-id")
```
ooq, line 29: Should test coverage here also includes other key types?

pkg/kritis/cryptolib/verifier_test.go
```diff
@@ -0,0 +1,75 @@
...
+	authenticatedAttestationChecker = func(authenticatedAttestation, string) error { return nil }
+
+	tcs := []struct {
+		name        string
+		att         *Attestation
+		publicKeys  []PublicKey
+		expectedErr bool
+	}{
+		{
+			name:        "happy case",
```
ooq, line 43: Test name needs to be more descriptive, e.g., "single key match"

pkg/kritis/cryptolib/verifier_test.go
```diff
@@ -0,0 +1,75 @@
...
+	nonmatchingKey := NewPublicKey(Pkix, []byte("key-data-other"), "key-id-other")
+
+	pkixVerify = func([]byte, []byte, []byte) error { return nil }
+	authenticatedAttestationChecker = func(authenticatedAttestation, string) error { return nil }
+
+	tcs := []struct {
+		name        string
+		att         *Attestation
+		publicKeys  []PublicKey
+		expectedErr bool
```
ooq, line 40: NIT: check error message

pkg/kritis/cryptolib/verifier_test.go
```diff
@@ -0,0 +1,75 @@
...
+	att := &Attestation{
+		PublicKeyID:       "key-id",
+		Signature:         []byte("signature"),
+		SerializedPayload: []byte("payload"),
+	}
+	matchingKey := NewPublicKey(Pkix, []byte("key-data"), "key-id")
+	otherMatchingKey := NewPublicKey(Pkix, []byte("key-data-other"), "key-id")
+	nonmatchingKey := NewPublicKey(Pkix, []byte("key-data-other"), "key-id-other")
+
+	pkixVerify = func([]byte, []byte, []byte) error { return nil }
```
ooq, line 33: Tests should also cover if `pkixVerify` returns error. (probably done in another testset.

pkg/kritis/cryptolib/verifier.go

pkg/kritis/cryptolib/verifier.go

pkg/kritis/cryptolib/verifier_test.go
acamadeo: Done.

pkg/kritis/cryptolib/verifier.go
acamadeo: I tried following this guidance. Please take a look to make sure I followed your suggestions appropriately.

pkg/kritis/cryptolib/verifier.go
acamadeo: Removed. Functions are no longer being globally stubbed.

pkg/kritis/cryptolib/verifier.go

pkg/kritis/cryptolib/verifier.go
acamadeo: Removed. Not necessary because public keys are now indexed at construction time. Finding is now a simple map lookup.

pkg/kritis/cryptolib/verifier_test.go
acamadeo: It probably should. To be clear, all the key-type specific verification tests will be happening in other test files (e.g. pgp_test.go). The tests here should only test high-level behavior of the `VerifyAttestation` method (i.e., key lookup, routing to the correct verify method).

For this test, I only included tests for key lookup. Does it also make sense to add tests for routing to the correct verify methods in this PR?

pkg/kritis/cryptolib/verifier_test.go
acamadeo: Done.

pkg/kritis/cryptolib/verifier_test.go
acamadeo: Acknowledged. Following Google's Go testing guidelines, which discourages string comparison of error messages, as those lead to change detector tests.

pkg/kritis/cryptolib/verifier_test.go
acamadeo: Done.

pkg/kritis/cryptolib/verifier.go
```diff
@@ -61,32 +89,42 @@ type verifier struct {
...
-	var (
-		err     error
-		payload []byte
-	)
-
 	// Extract the public key from `publicKeySet` whose ID matches the one in
 	// `att`.
-	// TODO: Replace no-op with correct implementation.
-	publicKey := v.PublicKeySet[0]
+	publicKey, found := v.PublicKeys[att.PublicKeyID]
```
alexcope, line 113: foo, ok := map[key] is idiomatic naming

pkg/kritis/cryptolib/verifier.go
```diff
@@ -61,32 +89,42 @@ type verifier struct {
...
+	if !found {
+		return fmt.Errorf("no public key with ID %s found", att.PublicKeyID)
```
alexcope, line 115: %q

pkg/kritis/cryptolib/verifier_test.go
```diff
@@ -0,0 +1,112 @@
...
+			name:         "error in verification",
+			att:          att,
+			publicKeys:   []PublicKey{matchingKey},
+			pkixVerifier: testVerifierFail{},
+			expectedErr:  true,
+		},
+	}
+	for _, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			v, err := NewVerifier("test-image-digest", tc.publicKeys)
```
alexcope, line 81: Instead of calling NewVerifier and casting, manually construct the internal verifier struct. tc.publicKeys can be a map or you can factor out an internal slice to map function that NewVerifier and the tests both call

pkg/kritis/cryptolib/verifier_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+			err = v.VerifyAttestation(tc.att)
+			if tc.expectedErr != (err != nil) {
+				t.Errorf("VerifyAttestation(_) got %v, wanted error? = %v", err, tc.expectedErr)
+			}
+		})
+	}
+}
+
+type testVerifierPass struct{}
```
alexcope, line 98: mockPkixVerifierAlwaysPass

pkg/kritis/cryptolib/pgp_test.go
```diff
@@ -118,7 +118,8 @@ func TestVerifyPgp(t *testing.T) {
 
 	for _, tc := range tcs {
 		t.Run(tc.name, func(t *testing.T) {
-			actualPayload, err := verifyPgp(tc.signature, tc.publicKey)
+			v := actualPgpVerifier{}
```
alexcope, line 121: nit: move construction out of test loop

pkg/kritis/cryptolib/pgp.go
```diff
@@ -25,11 +25,13 @@ import (
 	"golang.org/x/crypto/openpgp/armor"
 )
 
+type actualPgpVerifier struct{}
```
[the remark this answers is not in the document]
acamadeo: Done.

pkg/kritis/cryptolib/verifier.go

pkg/kritis/cryptolib/verifier.go

pkg/kritis/cryptolib/verifier_test.go
acamadeo: Done.

pkg/kritis/cryptolib/verifier_test.go
acamadeo: Modified mockPkixVerifier to contain `shouldErr` value.

pkg/kritis/cryptolib/pgp_test.go
acamadeo: Done.

pkg/kritis/cryptolib/verifier_test.go
alexcope: That code is simple enough and I don't see an elegant way to test it so I think it's fine as is

ooq:
@acamadeo I see that the kokoro tests are failing. Did you confirm with local integration tests?

acamadeo:
Local integration tests have passed.

PR #500, opened 2020-05-01
Minimal integration of crypto interface into Kritis

acamadeo:
Modify `GetValidatedAttestations` to call the new crypto interface.
Issue: #501

### DIFF
--- a/pkg/kritis/review/review_test.go
+++ b/pkg/kritis/review/review_test.go
@@ -222,13 +222,15 @@ func TestReviewGAP(t *testing.T) {
 			hasRequiredAtts: false,
 			shouldErr:       false,
 		},
+		// TODO: After payload check implemented in cryptolib,
+		// `hasRequiredAtts` should be false, and `shouldErr` should be true.
 		{
 			name:            "image with invalid attestation",
 			image:           img,
 			policies:        oneGAP,
 			attestations:    invalidAtts,
-			hasRequiredAtts: false,
-			shouldErr:       true,
+			hasRequiredAtts: true,
+			shouldErr:       false,
 		},
 		{
 			name:            "image complies with one policy out of two",
@@ -280,7 +282,7 @@ func TestReviewGAP(t *testing.T) {
 				Strategy:  &th,
 			})
 			if err := r.ReviewGAP([]string{tc.image}, tc.policies, nil, cMock); (err != nil) != tc.shouldErr {
-				t.Errorf("expected review to return error %t, actual error %s", tc.shouldErr, err)
+				t.Errorf("expected review to return error %t, actual error %v", tc.shouldErr, err)
 			}
 			if th.Attestations[tc.image] != tc.hasRequiredAtts {
 				t.Errorf("expected to have all required attestations for the image: %t. Got %t", tc.hasRequiredAtts, th.Attestations[tc.image])
--- a/pkg/kritis/review/review_test.go
+++ b/pkg/kritis/review/review_test.go
@@ -222,15 +222,13 @@ func TestReviewGAP(t *testing.T) {
 			hasRequiredAtts: false,
 			shouldErr:       false,
 		},
-		// TODO: After payload check implemented in cryptolib,
-		// `hasRequiredAtts` should be false, and `shouldErr` should be true.
 		{
 			name:            "image with invalid attestation",
 			image:           img,
 			policies:        oneGAP,
 			attestations:    invalidAtts,
-			hasRequiredAtts: true,
-			shouldErr:       false,
+			hasRequiredAtts: false,
+			shouldErr:       true,
 		},
 		{
 			name:            "image complies with one policy out of two",
@@ -282,7 +280,7 @@ func TestReviewGAP(t *testing.T) {
 				Strategy:  &th,
 			})
 			if err := r.ReviewGAP([]string{tc.image}, tc.policies, nil, cMock); (err != nil) != tc.shouldErr {
-				t.Errorf("expected review to return error %t, actual error %v", tc.shouldErr, err)
+				t.Errorf("expected review to return error %t, actual error %s", tc.shouldErr, err)
 			}
 			if th.Attestations[tc.image] != tc.hasRequiredAtts {
 				t.Errorf("expected to have all required attestations for the image: %t. Got %t", tc.hasRequiredAtts, th.Attestations[tc.image])
--- a/pkg/kritis/review/validating_transport.go
+++ b/pkg/kritis/review/validating_transport.go
@@ -102,26 +102,34 @@ func (avt *AttestorValidatingTransport) validatePublicKeyId(pubKey v1beta1.Publi
 	return nil
 }
 
-func (avt *AttestorValidatingTransport) parsePublicKeys() []cryptolib.PublicKey {
+// ParsePublicKeys fetches public keys from the Attestor spec. It returns the
+// valid PublicKeys and the keyID's for keys that could not be parsed.
+func (avt *AttestorValidatingTransport) parsePublicKeys() ([]cryptolib.PublicKey, []string) {
 	attestorKeys := avt.Attestor.Spec.PublicKeys
 	numKeys := len(attestorKeys)
-	publicKeys := []cryptolib.PublicKey{}
+	publicKeys, invalidKeys := []cryptolib.PublicKey{}, []string{}
 
 	for i, attestorKey := range attestorKeys {
+		if attestorKey.KeyType != "PGP" {
+			glog.Warningf("Skipping key %q with unsupported type %q", attestorKey.KeyId, attestorKey.KeyType)
+			continue
+		}
 		if err := avt.validatePublicKey(attestorKey); err != nil {
-			// warning level because single key failure is something tolerable
+			// Warning level because single key failure is something tolerable
 			glog.Warningf("Error parsing key %d (%d keys total) for %q: %v", i, numKeys, avt.Attestor.Name, err)
+			invalidKeys = append(invalidKeys, attestorKey.KeyId)
 			continue
 		}
 		decodedKey, err := base64.StdEncoding.DecodeString(attestorKey.AsciiArmoredPgpPublicKey)
 		if err != nil {
-			glog.Infof("Cannot base64 decode public key: %v", err)
+			glog.Warningf("Cannot base64 decode public key: %v", err)
+			invalidKeys = append(invalidKeys, attestorKey.KeyId)
 			continue
 		}
 		publicKey := cryptolib.NewPublicKey(cryptolib.Pgp, decodedKey, attestorKey.KeyId)
 		publicKeys = append(publicKeys, publicKey)
 	}
-	return publicKeys
+	return publicKeys, invalidKeys
 }
 
 func (avt *AttestorValidatingTransport) fetchAttestations(image string) ([]*cryptolib.Attestation, error) {
@@ -133,8 +141,7 @@ func (avt *AttestorValidatingTransport) fetchAttestations(image string) ([]*cryp
 
 	for _, rawAtt := range rawAtts {
 		if rawAtt.SignatureType != metadata.PgpSignatureType {
-			glog.Warningf("Skipping attestation with unsupported signature type %s", rawAtt.SignatureType.String())
-			continue
+			return nil, fmt.Errorf("Signature type %s is not supported for Attestation %v", rawAtt.SignatureType.String(), rawAtt)
 		}
 		decodedSig, err := base64.StdEncoding.DecodeString(rawAtt.Signature.Signature)
 		if err != nil {
@@ -154,9 +161,9 @@ func (avt *AttestorValidatingTransport) fetchAttestations(image string) ([]*cryp
 }
 
 func (avt *AttestorValidatingTransport) GetValidatedAttestations(image string) ([]attestation.ValidatedAttestation, error) {
-	publicKeys := avt.parsePublicKeys()
+	publicKeys, invalidKeys := avt.parsePublicKeys()
 	if len(publicKeys) == 0 {
-		return nil, fmt.Errorf("unable to find any valid key for %q", avt.Attestor.Name)
+		return nil, fmt.Errorf("Unable to find any valid keys for %q. Unparseable keys: %v", avt.Attestor.Name, invalidKeys)
 	}
 
 	verifier, err := cryptolib.NewVerifier(image, publicKeys)
--- a/pkg/kritis/review/validating_transport.go
+++ b/pkg/kritis/review/validating_transport.go
@@ -22,11 +22,13 @@ import (
 	"net/url"
 
 	"github.com/golang/glog"
+	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/grafeas/kritis/pkg/kritis/apis/kritis/v1beta1"
 	"github.com/grafeas/kritis/pkg/kritis/attestation"
-	"github.com/grafeas/kritis/pkg/kritis/container"
+	"github.com/grafeas/kritis/pkg/kritis/cryptolib"
 	"github.com/grafeas/kritis/pkg/kritis/metadata"
 	"github.com/grafeas/kritis/pkg/kritis/secrets"
+	"github.com/pkg/errors"
 )
 
 // ValidatingTransport allows the caller to obtain validated attestations for a given container image.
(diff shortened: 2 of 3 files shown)


pkg/kritis/review/validating_transport.go
```diff
@@ -101,51 +103,74 @@ func (avt *AttestorValidatingTransport) validatePublicKeyId(pubKey v1beta1.Publi
 	return nil
 }
 
+// NOTE: I am considering changing this function signature to:
```
nenaddedic, line 106: I advise against doing this.  In the bright future of attestations containing payloads, the caller of this function will be interested in the payload.

pkg/kritis/review/validating_transport.go
```diff
@@ -101,51 +103,74 @@ func (avt *AttestorValidatingTransport) validatePublicKeyId(pubKey v1beta1.Publi
...
+// verifier.VerifyAttestation call. It currently returns ValdiatedAttestations,
+// which contain unused and not particularly meaningful information.
+// This is not a trivial change, and would be done in a follow-up PR.
 func (avt *AttestorValidatingTransport) GetValidatedAttestations(image string) ([]attestation.ValidatedAttestation, error) {
-	keys := map[string]v1beta1.PublicKey{}
+	publicKeys := []cryptolib.PublicKey{}
 	numKeys := len(avt.Attestor.Spec.PublicKeys)
-	for i, pubKey := range avt.Attestor.Spec.PublicKeys {
-		if err := avt.validatePublicKey(pubKey); err != nil {
+	validatedAtts := []attestation.ValidatedAttestation{}
```
nenaddedic, line 115: Please move to L150 or so, closer to the place where it is first needed and used.

pkg/kritis/review/validating_transport.go
```diff
@@ -101,51 +103,74 @@ func (avt *AttestorValidatingTransport) validatePublicKeyId(pubKey v1beta1.Publi
...
+
+	for i, attestorKey := range avt.Attestor.Spec.PublicKeys {
```
nenaddedic, line 117: I suggest factoring this into a helper function "parsePublicKeys" or something such.

pkg/kritis/review/validating_transport.go
```diff
@@ -101,51 +103,74 @@ func (avt *AttestorValidatingTransport) validatePublicKeyId(pubKey v1beta1.Publi
...
-	rawAtts, err := avt.Client.Attestations(image, &avt.Attestor)
+
+	verifier, err := cryptolib.NewVerifier(digest.DigestStr(), publicKeys)
 	if err != nil {
-		glog.Error(err)
-		return nil, err
+		return nil, errors.Wrap(err, "error creating verifier")
 	}
 
+	rawAtts, err := avt.Client.Attestations(image, &avt.Attestor)
```
nenaddedic, line 147: Do we plan to change this to return cryptolib.Attestation instead of RawAttestation?  Would that be a good idea?

(No need to do it now, I'm just curious.)

nenaddedic, line 147: Also, this may be easier to read if you factor out the fetch-attestations-and-convert into a helper function:
func fetchAttestations(...) ([]cryptolib.Attestation, error)

and then do the verification in a loop after that

pkg/kritis/review/validating_transport.go
```diff
@@ -101,51 +103,74 @@ func (avt *AttestorValidatingTransport) validatePublicKeyId(pubKey v1beta1.Publi
...
+
+	for i, attestorKey := range avt.Attestor.Spec.PublicKeys {
+		if err := avt.validatePublicKey(attestorKey); err != nil {
 			// warning level because single key failure is something tolerable
 			glog.Warningf("Error parsing key %d (%d keys total) for %q: %v", i, numKeys, avt.Attestor.Name, err)
 			continue
 		}
-		if _, ok := keys[pubKey.KeyId]; ok {
-			glog.Warningf("Overwriting key with same fingerprint %s for %q.", pubKey.KeyId, avt.Attestor.Name)
+		// Base64 decode the attestor's key
```
alexcope, line 123: nit: remove this comment because it does not explain anything that is not obvious from the code

pkg/kritis/review/validating_transport.go
```diff
@@ -101,51 +103,74 @@ func (avt *AttestorValidatingTransport) validatePublicKeyId(pubKey v1beta1.Publi
...
+		decodedKey, err := base64.StdEncoding.DecodeString(attestorKey.AsciiArmoredPgpPublicKey)
```
alexcope, line 124: Do we really use standard encoding instead of web-safe base64 encoding?

pkg/kritis/review/review_test.go
```diff
@@ -222,13 +222,15 @@ func TestReviewGAP(t *testing.T) {
 			hasRequiredAtts: false,
 			shouldErr:       false,
 		},
+		// TODO: After payload check implemented in cryptolib,
+		// `hasRequiredAtts` should be false, and `shouldErr` should be true.
```
alexcope, line 226: What is this trying to test right now? I'd rather disable a test that does not work because something is unimpliemented than have a test that expects incorrect behaviour

pkg/kritis/cryptolib/verifier.go
```diff
@@ -123,12 +143,10 @@ func formAuthenticatedAttestation(payload []byte) authenticatedAttestation {
 	return authenticatedAttestation{}
 }
 
-// Check that the data within the Attestation payload matches what we expect.
+// TODO: Check that the data within the Attestation payload matches what we
+// expect.
 // NOTE: This is a simple comparison for plain attestations, but it would be
 // more complex for rich attestations.
 func checkAuthenticatedAttestation(actual authenticatedAttestation, imageDigest string) error {
-	if actual.ImageDigest != imageDigest {
```
ooq, line 130: Why are these deleted?

pkg/kritis/review/review_test.go
```diff
@@ -222,13 +222,15 @@ func TestReviewGAP(t *testing.T) {
...
 		{
 			name:            "image with invalid attestation",
 			image:           img,
 			policies:        oneGAP,
 			attestations:    invalidAtts,
-			hasRequiredAtts: false,
-			shouldErr:       true,
+			hasRequiredAtts: true,
```
ooq, line 232: Can we update the PR so that it doesn't change existing test behavior?

pkg/kritis/review/validating_transport.go
```diff
@@ -101,51 +103,74 @@ func (avt *AttestorValidatingTransport) validatePublicKeyId(pubKey v1beta1.Publi
...
+	rawAtts, err := avt.Client.Attestations(image, &avt.Attestor)
+	if err != nil {
+		return nil, fmt.Errorf("error fetching attestations for image %s: %v", image, err)
+	}
 	for _, rawAtt := range rawAtts {
-		if rawAtt.SignatureType != metadata.PgpSignatureType {
-			return nil, fmt.Errorf("Signature type %s is not supported for Attestation %v", rawAtt.SignatureType.String(), rawAtt)
+		if rawAtt.SignatureType != metadata.PgpSignatureType && rawAtt.SignatureType != metadata.GenericSignatureType {
+			glog.Warningf("Skipping attestation with unsupported signature type %s", rawAtt.SignatureType.String())
+			continue
```
ooq, line 154: To silently pass unsupported attestation type can potentially makes it hard to debug. That was the reason for failing it previously
If we intend to change this behavior,  it probably should be done in a separate PR.

pkg/kritis/review/validating_transport.go
ooq: We (me & @acamadeo ) discussed on this previously.
I think eventually `RawAttestation` could be completely replaced by `cryptolib.Attestation`. But due to `RawAttestation`'s large exposure surface, we'll for now do conversion between them and investigate replacement in later PRs.

pkg/kritis/review/validating_transport.go
```diff
@@ -101,51 +103,74 @@ func (avt *AttestorValidatingTransport) validatePublicKeyId(pubKey v1beta1.Publi
...
-			glog.Warningf("Could not find or verify attestation for attestor %s: %s", keyId, err.Error())
+		// TODO: Remove this after cryptolib.Attestation is used throughout
+		// Kritis.
+		att := &cryptolib.Attestation{
+			PublicKeyID:       rawAtt.Signature.PublicKeyId,
+			Signature:         decodedSig,
+			SerializedPayload: rawAtt.SerializedPayload,
+		}
+
+		if err := verifier.VerifyAttestation(att); err != nil {
```
ooq, line 169: Note that `verifier.VerifyAttestation` is actually not equivalent to `host.VerifySignature`. The latter also checks `Repository` I think. Does this not deprecate some supported cases?

pkg/kritis/review/validating_transport_test.go
```diff
@@ -237,29 +237,35 @@ func TestValidatingTransport(t *testing.T) {
 		{name: "invalid secret", auth: validAuthWithOneGoodPgpKey, wantAtts: []attestation.ValidatedAttestation{}, attestations: []metadata.RawAttestation{
 			metadata.MakeRawAttestation(metadata.PgpSignatureType, encodeB64("invalid-sig"), "invalid-fpr", ""),
 		}, errorExpected: false, attError: nil},
-		{name: "valid sig over another host", auth: validAuthWithOneGoodPgpKey, wantAtts: []attestation.ValidatedAttestation{}, attestations: []metadata.RawAttestation{
+		// TODO(acamadeo): After payload check is implemented in cryptolib,
+		// `wantAtts` should be an empty list.
+		{name: "valid sig over another host", auth: validAuthWithOneGoodPgpKey, wantAtts: []attestation.ValidatedAttestation{
```
ooq, line 242: This actually changes the behavior of Kritis. Can we avoid it?

pkg/kritis/review/validating_transport_test.go
```diff
@@ -237,29 +237,35 @@ func TestValidatingTransport(t *testing.T) {
...
 		{name: "attestation fetch error", auth: validAuthWithOneGoodPgpKey, wantAtts: nil, attestations: nil, errorExpected: true, attError: errors.New("can't fetch attestations")},
 		{name: "auth with invalid PGP key", auth: invalidAuthWithOneBadPgpKey, wantAtts: nil, attestations: []metadata.RawAttestation{
 			metadata.MakeRawAttestation(metadata.PgpSignatureType, encodeB64(sig), successFpr, ""),
 		}, errorExpected: true, attError: nil},
-		{name: "auth with generic signature type", auth: validAuthWithOneGoodPgpKey, wantAtts: nil, attestations: []metadata.RawAttestation{
+		{name: "auth with generic signature type", auth: validAuthWithOneGoodPgpKey, wantAtts: []attestation.ValidatedAttestation{}, attestations: []metadata.RawAttestation{
 			metadata.MakeRawAttestation(metadata.GenericSignatureType, "test-sig", "test-id", "generic-address"),
-		}, errorExpected: true, attError: nil},
-		{name: "auth with unknown signature type", auth: validAuthWithOneGoodPgpKey, wantAtts: nil, attestations: []metadata.RawAttestation{
+		}, errorExpected: false, attError: nil},
```
ooq, line 256: Ditto as above.

pkg/kritis/review/validating_transport_test.go
```diff
@@ -237,29 +237,35 @@ func TestValidatingTransport(t *testing.T) {
...
+		{name: "auth with unknown signature type", auth: validAuthWithOneGoodPgpKey, wantAtts: []attestation.ValidatedAttestation{}, attestations: []metadata.RawAttestation{
 			metadata.MakeRawAttestation(metadata.UnknownSignatureType, encodeB64(sig), successFpr, ""),
-		}, errorExpected: true, attError: nil},
+		}, errorExpected: false, attError: nil},
```
ooq, line 259: Ditto as above.

pkg/kritis/review/validating_transport_test.go
```diff
@@ -237,29 +237,35 @@ func TestValidatingTransport(t *testing.T) {
...
 		// TODO(acamadeo): After PKIX key verification implementation, the
 		// `wantAtts` field for this test case should be a list of
-		// ValidatedAttestations. `errorExpected` should also be false, but
-		// currently attestations with GenericSignatureType throw an error.
-		{name: "auth with valid PKIX key", auth: validAuthWithOneGoodPkixKey, wantAtts: nil, attestations: []metadata.RawAttestation{
+		// ValidatedAttestations.
+		{name: "auth with valid PKIX key", auth: validAuthWithOneGoodPkixKey, wantAtts: []attestation.ValidatedAttestation{}, attestations: []metadata.RawAttestation{
 			metadata.MakeRawAttestation(metadata.GenericSignatureType, "", "", ""),
-		}, errorExpected: true, attError: nil},
+		}, errorExpected: false, attError: nil},
```
ooq, line 268: Ditto as above.

pkg/kritis/review/validating_transport.go
acamadeo: Even if 1 Attestation has an unsupported type, wouldn't you still want to check the other Attestations retrieved by the client?

pkg/kritis/review/validating_transport.go
acamadeo: Done.

pkg/kritis/review/validating_transport.go
acamadeo: Done.

pkg/kritis/review/validating_transport.go
acamadeo: Yes, we will replace `RawAttestation` with `cryptolib.Attestation` in a future PR

acamadeo: Done.

pkg/kritis/review/validating_transport.go
acamadeo: Done.

pkg/kritis/review/review_test.go
acamadeo: Reverted to original behavior.

pkg/kritis/review/review_test.go
acamadeo: Reverted to original behavior.

pkg/kritis/cryptolib/verifier.go
acamadeo: This check is now implemented in #502.

pkg/kritis/review/validating_transport.go
acamadeo: Incorporated repository check in #502

pkg/kritis/review/validating_transport_test.go
acamadeo: Reverted to original behavior.

pkg/kritis/review/validating_transport_test.go
acamadeo: GenericSignatureType is now supported via the crypto library, so this should not fail.

pkg/kritis/review/validating_transport_test.go
acamadeo: Attestations with UnknownSignatureType are now skipped over, so they should produce no error

pkg/kritis/review/validating_transport_test.go
acamadeo: Reverted to original behavior.

pkg/kritis/review/validating_transport.go
acamadeo: We currently use StdEncoding in Kritis. We should discuss whether we want to migrate to UrlEncoding (https://godoc.org/encoding/base64#pkg-variables). This change would be out of scope for this PR.

pkg/kritis/review/validating_transport.go
```diff
@@ -101,51 +102,80 @@ func (avt *AttestorValidatingTransport) validatePublicKeyId(pubKey v1beta1.Publi
...
+			Signature:         decodedSig,
+			SerializedPayload: rawAtt.SerializedPayload,
+		}
+		atts = append(atts, att)
+	}
+	return atts, nil
+}
+
+func (avt *AttestorValidatingTransport) GetValidatedAttestations(image string) ([]attestation.ValidatedAttestation, error) {
+	publicKeys := avt.parsePublicKeys()
```
nenaddedic, line 157: Consider changing parsePublicKeys() to return (keys, error).  There can be many errors, and any aggregation will work.  For example, just send back the first or the last one encountered.  Or even just ids of offending keys.

If you do this, then parsing errors won't be entirely swallowed, and it will be easier for user to debug.

pkg/kritis/review/validating_transport.go
ooq: If we are changing behavior of kritis, we should probably do it in a separate PR.

pkg/kritis/review/validating_transport.go
ooq: I don't have much knowledge on the subject. What's the big diff here? Should we open an issue tracking this?

pkg/kritis/review/validating_transport.go
ooq: ditto on returning parsing errors.

I am not sure if we should pass error when there is at least one valid key, but definitely aggregate errors when no key is valid.

pkg/kritis/review/validating_transport.go
```diff
@@ -101,51 +102,80 @@ func (avt *AttestorValidatingTransport) validatePublicKeyId(pubKey v1beta1.Publi
...
-	numKeys := len(avt.Attestor.Spec.PublicKeys)
-	for i, pubKey := range avt.Attestor.Spec.PublicKeys {
-		if err := avt.validatePublicKey(pubKey); err != nil {
+func (avt *AttestorValidatingTransport) parsePublicKeys() []cryptolib.PublicKey {
+	attestorKeys := avt.Attestor.Spec.PublicKeys
+	numKeys := len(attestorKeys)
+	publicKeys := []cryptolib.PublicKey{}
+
+	for i, attestorKey := range attestorKeys {
+		if err := avt.validatePublicKey(attestorKey); err != nil {
```
ooq, line 111: I might miss something, should we check keytype here?

pkg/kritis/review/validating_transport.go
acamadeo: Done. Reverted

pkg/kritis/review/validating_transport_test.go
acamadeo: Reverted, as discussed above.

pkg/kritis/review/validating_transport.go
acamadeo: I modified this to also return the key ID's for keys that could not be parsed. The caller will error if none of the keys were valid, and its error message references the ID's of the invalid keys. The specific error for each invalid key should be accessible from the logs.

pkg/kritis/review/validating_transport.go
acamadeo: Done, added. Perhaps we can remove this check once PKIX verification is implemented.

ooq: Sure. I'm bringing up `check` mostly because we call `attestorKey.AsciiArmoredPgpPublicKey` later in this function, not sure why I left the comment here...

pkg/kritis/review/validating_transport.go
```diff
@@ -101,34 +102,41 @@ func (avt *AttestorValidatingTransport) validatePublicKeyId(pubKey v1beta1.Publi
...
+			continue
 		}
-		keys[pubKey.KeyId] = pubKey
-	}
-	if len(keys) == 0 {
-		return nil, fmt.Errorf("unable to find any valid key for %q", avt.Attestor.Name)
+		publicKey := cryptolib.NewPublicKey(cryptolib.Pgp, decodedKey, attestorKey.KeyId)
+		publicKeys = append(publicKeys, publicKey)
 	}
+	return publicKeys, invalidKeys
```
ooq, line 132: NIT: would still be good to explicitly return error if len(publicKeys) == 0. How it is structured now relies on the caller function to do things properly.

pkg/kritis/review/validating_transport_test.go
```diff
@@ -255,8 +255,7 @@ func TestValidatingTransport(t *testing.T) {
 		}, errorExpected: false, attError: nil},
 		// TODO(acamadeo): After PKIX key verification implementation, the
 		// `wantAtts` field for this test case should be a list of
-		// ValidatedAttestations. `errorExpected` should also be false, but
-		// currently attestations with GenericSignatureType throw an error.
+		// ValidatedAttestations.
```
ooq, line 258: Just make sure it's intended update.

pkg/kritis/review/validating_transport.go
acamadeo: We decided offline that `parsePublicKeys` should not return an error, but should return successful and failed parsing results. We decided to provide the attestor's public keys as an argument.

pkg/kritis/review/validating_transport_test.go
acamadeo: Done.

acamadeo:

PR #504, opened 2020-05-07
Correct the LICENSE text.

nenaddedic:


LICENSE
```diff
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2018 Google LLC
+   Copyright [yyyy] [name of copyright owner]
```
aysylu, line 190: Is this intentional, to remove the year and company?

nenaddedic: Yes.  Note that this is in the appendix starting at L179, and is part of the cut-and-paste template that one can use to apply the license to their work.  IOW, this part of the text applies to the license itself, not to Kritis.

aysylu: Oh I see now, thank you for clarifying.

aysylu:
The failing kokoro test is unrelated to the PR.

PR #511, opened 2020-05-18
[ISSUE-506] Remove base64 encoding/decoding for attestation signatures.

### LINKED ISSUE #506: BinAuthz/Kritis incompatibility: base64-encoded attestation
## Expected Behavior
The same attestation should work for both Kritis & Binary Authorization.

## Actual Behavior
Kritis assumes signature in attestation to be base64 encoded. (and Kritis also creates attestation with base64 encoded signature)
BinAuthz assumes not.

## Steps to Reproduce the Problem

Environment, commands

1.  
2.  
3.  
4.  

## Additional info
Add any other context about the problem here (e.g. screenshots, links)

ooq:
Fix #506 .


pkg/kritis/attestation/attestation_test.go
```diff
@@ -61,8 +61,8 @@ func TestGPGArmorSignIntegration(t *testing.T) {
 	}
 }
 
-// Base64 encoded signature.
-// Created using echo  test >  test && gpg --armor --sign -u test@kritis.org < test | base64
+//  signature.
+// Created using echo  test >  test && gpg --armor --sign -u test@kritis.org < test
```
ooq, line 65: Comment: no change to sig content because the old comment was wrong.

pkg/kritis/container/container_test.go
```diff
@@ -214,7 +214,7 @@ func TestContainerEquals(t *testing.T) {
 	}
 }
 
-// Created using gpg --armor --sign -u test@kritis.org <atomic_host_json_representation.txt> | base64
+// Created using gpg --armor --sign -u test@kritis.org <atomic_host_json_representation.txt>
```
ooq, line 217: Comment: no change to sig content because the old comment was wrong.

pkg/kritis/testutil/constants.go
```diff
@@ -22,7 +22,7 @@ const (
 	PgpKeyFingerprint = "D283A5F5F5F5ECAA9EF185C5AE8B6994116315A3"
 )
 
-// PublicTestKey is base64 encoded sig for image gcr.io/kritis-project/kritis-server@sha256:b3f3eccfd27c9864312af3796067e7db28007a1566e1e042c5862eed3ff1b2c8
+// PublicTestKey is a base64 encoded public key (base64-encoded on top of the base64-encoded ASCII-armored format).
```
ooq, line 25: Fix a comment not directly related to sig format change.

nenaddedic:
I think this PR is correct, but I wonder if the issue described in https://groups.google.com/g/kritis-users/c/atVVPZmzs24/m/xyb-oKvYAAAJ would now re-occur.  Have you verified this by any chance?

ooq:
@nenaddedic 
The issue in  https://groups.google.com/g/kritis-users/c/atVVPZmzs24/m/xyb-oKvYAAAJ should be fixed by not base64 signature field, or using the v1 API instead, basically:

Signature field in attestation is a string type in [v1beta1 API](https://github.com/grafeas/grafeas/blob/master/proto/v1beta1/attestation.proto#L39), so it should not be base64 encoded.
Signature field is bytes type in [v1 API](https://github.com/grafeas/grafeas/blob/master/proto/v1/common.proto#L83), and it should be base64 encoded before putting into a json struct.

ba32107:
Hello,

I was testing Kritis for the past couple of days, and this change suddenly broke our test setup. I've [opened a PR](https://github.com/grafeas/kritis/pull/514) to fix the documentation.

On a related note, is there any communications channel where new container image versions and their changelogs are announced? I was using the `latest` tag, which is a mistake I fixed now, but I'd like to be kept up-to-date on the new releases and breaking changes.

Thanks,
Balazs

ooq:
Hi @ba32107 , what container images are you referring to?

ba32107:
Hi @ooq ,

I was referring to the [container registry](https://console.cloud.google.com/gcr/images/kritis-project/GLOBAL/kritis-server?gcrImageListsize=30) for Kritis. I used the image with the `latest` tag for my testing last week, and when I tried again today, `latest` was already pointing to the new version, which broke my test setup (because I was base64 encoding the signatures).

Using the `latest` tag is not a good idea and I've already fixed that, but I'd be interested to learn how can I keep up-to-date with the newest Kritis releases and their changelogs (the [changelog file](https://github.com/grafeas/kritis/blob/master/CHANGELOG.md) seems to be out of date).

Thanks

ooq:
AFAIK, those images are for testing purposes only. @aysylu , is this correct?

ba32107:
Ah that may be, in that case my mistake :)

aysylu:
> @nenaddedic
> The issue in https://groups.google.com/g/kritis-users/c/atVVPZmzs24/m/xyb-oKvYAAAJ should be fixed by not base64 signature field, or using the v1 API instead, basically:
> 
> Signature field in attestation is a string type in [v1beta1 API](https://github.com/grafeas/grafeas/blob/master/proto/v1beta1/attestation.proto#L39), so it should not be base64 encoded.
> Signature field is bytes type in [v1 API](https://github.com/grafeas/grafeas/blob/master/proto/v1/common.proto#L83), and it should be base64 encoded before putting into a json struct.

@ooq since v1 API isn't available in the Kritis implementation (and the migration time is TBD), could you please update documentation to guide the users with this recommendation, to avoid running into the same issue again? Thanks! Happy to review the PR.:)

aysylu:
> Hello,
> 
> I was testing Kritis for the past couple of days, and this change suddenly broke our test setup. I've [opened a PR](https://github.com/grafeas/kritis/pull/514) to fix the documentation.
> 
> On a related note, is there any communications channel where new container image versions and their changelogs are announced? I was using the `latest` tag, which is a mistake I fixed now, but I'd like to be kept up-to-date on the new releases and breaking changes.
> 
> Thanks,
> Balazs

@ba32107: yes, please join [kritis-users](https://groups.google.com/g/kritis-users) mailing list to get announcements of releases of the new versions.

aysylu:
> AFAIK, those images are for testing purposes only. @aysylu , is this correct?

@ba32107 and @ooq: Qifan is correct AND we also store images from Kritis releases there. I'd recommend using the versioned tags instead, eg `v0.2.2`. Does this help?

ba32107:
> > AFAIK, those images are for testing purposes only. @aysylu , is this correct?
> 
> @ba32107 and @ooq: Qifan is correct AND we also store images from Kritis releases there. I'd recommend using the versioned tags instead, eg `v0.2.2`. These correspond to the Kritis releases we've had. Does this help?

Indeed it does, thank you @aysylu!

ba32107:
> @ba32107: yes, please join [kritis-users](https://groups.google.com/g/kritis-users) mailing list to get announcements of releases of the new versions.

Interestingly, this link does not work for me. I get a 404:
```
404. That’s an error.
The requested URL was not found on this server. That’s all we know.
```
However, the following link does work and I am part of the group:
https://groups.google.com/forum/#!forum/kritis-users

I presume these two URLs are pointing to the same place?

aysylu:
> > @ba32107: yes, please join [kritis-users](https://groups.google.com/g/kritis-users) mailing list to get announcements of releases of the new versions.
> 
> Interestingly, this link does not work for me. I get a 404:
> 
> ```
> 404. That’s an error.
> The requested URL was not found on this server. That’s all we know.
> ```
> 
> However, the following link does work and I am part of the group:
> https://groups.google.com/forum/#!forum/kritis-users
> 
> I presume these two URLs are pointing to the same place?

That's strange! And yes, it's the same URL.

ooq:
I think the `/g/` part or the url `https://groups.google.com/g/kritis-users` probably requires Google internal credentials to redirect correctly, we should avoid using such links in the future.

@aysylu In the standalone doc, we recommended using our provided script for creating attestation: https://github.com/grafeas/kritis/tree/master/docs/standalone
This should avoid the issue? I feel that this is quite an implementation nuance that is caused by API inconsistency, we might want to avoid prescribing it before the API stabilizes?

aysylu:
@ooq sounds good, yes, this should be sufficient. Thanks for checking!

PR #514, opened 2020-05-20
Update resources.md: remove base64 encoding

ba32107:
Base64 support was removed in https://github.com/grafeas/kritis/pull/511 - updating the documentation accordingly.

### DIFF
--- a/docs/resources.md
+++ b/docs/resources.md
@@ -267,7 +267,7 @@ an additional `keyType` field.
 |----------|----------|---------------|
 | `keyType`  | string   | "PGP"| 
 | `keyId`  | string   | Optional: either OpenPGP RFC4880 V4 fingerprint of the key payload or blank. If left blank, the `keyId` will be computed as the key's OpenPGP fingerprint.| 
-| `asciiArmoredPgpPublicKey`  | string | A base64-encoded ASCII-armored representation of a PGP public key, as the entire output by the command `gpg --export --armor foo@example.com | base64` (either LF or CRLF line endings). |
+| `asciiArmoredPgpPublicKey`  | string | An ASCII-armored string representation of a PGP public key, as the entire output by the command `gpg --export --armor foo@example.com | base64` (either LF or CRLF line endings). |
 | `pkixPublicKey` | PkixPublicKey object | empty|
 
 ### PKIX Public Keys


ooq:
Thanks for the PR @ba32107 .
We removed base64-encoding for attestation signature, but we still have base64 encoding for pgp keys.

ba32107:
Ah, okay you are completely right, my mistake. However I'm pretty sure I have read somewhere that the signature must also be base64 encoded. I checked the Grafeas proto files but I can't find the doc I am referring to.

Does this ring a bell for you? I wanted to fix _that_ documentation, but I'm not even sure it exists :)

ba32107:
Perhaps I inferred it from the code, I am not sure. This PR is not needed though.

ooq:
@ba32107 If there is something in kritis we say that signature needs to be base64 encoded, we should definitely fix.
In addition, I think it would be great to document the fix to this issue in Grafeas documentation [1]: basically the fix here should be to drop base64 encoding for grafeas v1beta1 API (because proto3 does automatic base64 decoding when converting json string to bytes field), and use base64 encoding for grafeas v1 API.

https://groups.google.com/g/kritis-users/c/atVVPZmzs24/m/xyb-oKvYAAAJ

ba32107:
Hi @ooq,

I did not find any documentation in Kritis nor Grafeas that mention having to base64 encode the signature, so I believe the documentation is correct. I must have figured out from the code that I needed to base64 encode. However, I did find a remaining reference to base64 decoding just now:
https://github.com/grafeas/kritis/blob/fca6c5f70519521784c042b74960ed455365a64b/pkg/kritis/review/validating_transport.go#L146
I believe that `if` block is not necessary and this was left out of #511.



> and use base64 encoding for grafeas v1 API

Could you clarify this part, please? I don't follow this. If the `v1` API has `bytes` for signature, why do we need to base64 encode? Isn't that something that the gRPC library takes care of for us? 

So to summarize, is my understanding correct here?
 * When using `v1beta1`, I should put the signature as an ASCII armored string into Grafeas (no encoding), and Kritis doesn't do any decoding either
 * When using `v1`, I should send a raw byte stream signature into Grafeas, and Kritis will also receive the same byte stream (once v1 is implemented in Kritis)

Is this correct?

> https://groups.google.com/g/kritis-users/c/atVVPZmzs24/m/xyb-oKvYAAAJ

On an unrelated note, this link does not work for me, even though I am part of the kritis-users google group - do I need permissions?

ooq:
> Hi @ooq,
> 
> I did not find any documentation in Kritis nor Grafeas that mention having to base64 encode the signature, so I believe the documentation is correct. I must have figured out from the code that I needed to base64 encode. However, I did find a remaining reference to base64 decoding just now:
> https://github.com/grafeas/kritis/blob/fca6c5f70519521784c042b74960ed455365a64b/pkg/kritis/review/validating_transport.go#L146
> 
> 
> I believe that `if` block is not necessary and this was left out of #511.
> > and use base64 encoding for grafeas v1 API
Nice catch! We should remove that error message. If you could open a PR to fix it, I should be able to quickly approve it.
> 
> Could you clarify this part, please? I don't follow this. If the `v1` API has `bytes` for signature, why do we need to base64 encode? Isn't that something that the gRPC library takes care of for us?
> 
You are right, I don't think you need to do anything with the client library. But if you do create attestation with curl commands and passing a json object -- like what's described in that google groups discussion. You need to do base64 encoding manually for v1 API.
> So to summarize, is my understanding correct here?
> 
> * When using `v1beta1`, I should put the signature as an ASCII armored string into Grafeas (no encoding), and Kritis doesn't do any decoding either
> * When using `v1`, I should send a raw byte stream signature into Grafeas, and Kritis will also receive the same byte stream (once v1 is implemented in Kritis)
> 
> Is this correct?
> 
> > https://groups.google.com/g/kritis-users/c/atVVPZmzs24/m/xyb-oKvYAAAJ
> 
> On an unrelated note, this link does not work for me, even though I am part of the kritis-users google group - do I need permissions?
Strange..you should be able to see it as long as you are in kritis-users. @aysylu , do you have any idea?

ba32107:
> But if you do create attestation with curl commands and passing a json object -- like what's described in that google groups discussion. You need to do base64 encoding manually for v1 API.

Understood, we are on the same page, thank you!

Opened https://github.com/grafeas/kritis/pull/515 to fix the mentioned issue.

PR #519, opened 2020-05-29
Replace RawAttestation with Cryptolib Attestation

acamadeo:
Replace all instances of `RawAttestation` with cryptolib `Attestation`. Includes documentation fixes for functions and structs.


pkg/kritis/metadata/containeranalysis/cache_test.go
```diff
@@ -139,12 +141,10 @@ func TestNCache(t *testing.T) {
...
 }
 
-func makeRawAttestationPgp(signature, id string) metadata.RawAttestation {
-	return metadata.RawAttestation{
-		SignatureType: metadata.PgpSignatureType,
-		Signature: metadata.RawSignature{
-			Signature:   signature,
-			PublicKeyId: id,
-		},
+func makePgpAttestation(signature, id string) cryptolib.Attestation {
```
alexcope, line 144: Nit: you didn't really add this function, but generally in tests objects like this one are constructed directly instead of using a helper function. Helper functions are only really used if the object is very complicated and only a few parameters change test to test.

pkg/kritis/metadata/metadata.go
```diff
@@ -130,49 +109,43 @@ func GetVulnerabilityFromOccurrence(occ *grafeas.Occurrence) *Vulnerability {
...
-	return RawAttestation{
-		SignatureType: sigType,
-		Signature: RawSignature{
-			Signature:   sig,
-			PublicKeyId: id,
-		},
-		SerializedPayload: []byte(payload),
+// MakeAttestation is testing purposes. Should not be used as part of metadata
+// external API.
+func MakeAttestation(publicKeyID string, signature string, serializedPayload []byte) cryptolib.Attestation {
```
alexcope, line 145: Not added in this CL, but think this kind of helper function is generally more harmful than beneficial. It's not that hard to directly construct cryptolib.Attestation.

Alternatively, move it to *_test.go and rename it to MakeTestAttestation. It's better to duplicate this function in multiple test files than to have iit here with the production code.

ooq:
Feel free to ignore the cloud build error for now.

pkg/kritis/metadata/containeranalysis/cache_test.go
acamadeo: Done, removed

pkg/kritis/metadata/metadata.go
acamadeo: Done, removed

pkg/kritis/metadata/metadata.go
```diff
@@ -130,49 +109,33 @@ func GetVulnerabilityFromOccurrence(occ *grafeas.Occurrence) *Vulnerability {
...
-				Signature:   psa.GetSignature(),
-			},
-			SerializedPayload: []byte{},
+		psa := occAtt.GetPgpSignedAttestation()
+		att := cryptolib.Attestation{
+			PublicKeyID: psa.GetPgpKeyId(),
+			Signature:   []byte(psa.GetSignature()),
 		}
-		ras = append(ras, ra)
+		atts = append(atts, att)
```
nenaddedic, line 126: I suggest returning []cryptolib.Attestation{att} directly.  Also below.

pkg/kritis/review/validating_transport.go
```diff
@@ -132,26 +132,11 @@ func (avt *AttestorValidatingTransport) parsePublicKeys(attestorKeys []v1beta1.P
 	return publicKeys, invalidKeys
 }
 
-func (avt *AttestorValidatingTransport) fetchAttestations(image string) ([]*cryptolib.Attestation, error) {
-	atts := []*cryptolib.Attestation{}
-	rawAtts, err := avt.Client.Attestations(image, &avt.Attestor)
+func (avt *AttestorValidatingTransport) fetchAttestations(image string) ([]cryptolib.Attestation, error) {
```
nenaddedic, line 135: In L154, you can just call avt.Client.Attestations directly and get rid of fetchAttestations() altogether.

PR #527, opened 2020-06-05
Support polling for vulnerability results in signer CLI

ooq:


cmd/kritis/signer/main.go
```diff
@@ -34,9 +35,10 @@ import (
 )
 
 func main() {
-	var image, json_key_path, pri_key_path, passphrase, pub_key_path, policy_path string
+	var image, timeout_str, json_key_path, pri_key_path, passphrase, pub_key_path, policy_path string
 
 	flag.StringVar(&image, "image", "", "image url, e.g., gcr.io/foo/bar@sha256:abcd")
+	flag.StringVar(&timeout_str, "timeout", "5m", "timeout for checking on image vulnerability , e.g., 600s, 5m")
```
nenaddedic, line 41: Please give it some more descriptive name.  containeranalysis_timeout, vulnz_timeout, or something else

pkg/kritis/metadata/containeranalysis/containeranalysis.go
```diff
@@ -294,3 +296,74 @@ func (c Client) DeleteOccurrence(ID string) error {
 	glog.Infof("executed deletion of occurrence=%s", ID)
 	return c.client.DeleteOccurrence(c.ctx, req)
 }
+
+// Poll discovery occurrence for an image and wait until container analysis
+// finishes. Throws an error if analysis is not successful or timeouts.
+func WaitForVulnzAnalysis(containerImage string, timeout time.Duration) error {
```
nenaddedic, line 302: Should this go to metadata.go?  It doesn't seem to be tied exclusively to the containeranalysis client.

Also, I suggest passing the client to the function, instead of creating it inside of it.

pkg/kritis/metadata/containeranalysis/containeranalysis.go
```diff
@@ -294,3 +296,74 @@ func (c Client) DeleteOccurrence(ID string) error {
...
+	defer cancel()
+
+	client, err := ca.NewGrafeasV1Beta1Client(ctx)
+	if err != nil {
+		return fmt.Errorf("NewClient: %v", err)
+	}
+	defer client.Close()
+
+	// ticker is used to poll once per second.
+	ticker := time.NewTicker(1 * time.Second)
```
nenaddedic, line 315: Some backoff may be a good idea.

pkg/kritis/metadata/containeranalysis/containeranalysis.go
```diff
@@ -294,3 +296,74 @@ func (c Client) DeleteOccurrence(ID string) error {
...
+			// Only one occurrence should ever be returned by ListOccurrences
+			// and the given filter.
+			result, err := it.Next()
+			if err == iterator.Done {
+				break
+			}
+			if err != nil {
+				return fmt.Errorf("it.Next: %v", err)
+			}
+			if result.GetDiscovered() != nil {
```
nenaddedic, line 341: Can you add a filtering clause to only consider Discovery occurrences?

pkg/kritis/metadata/containeranalysis/containeranalysis.go
```diff
@@ -294,3 +296,74 @@ func (c Client) DeleteOccurrence(ID string) error {
...
+				discoveryOccurrence = result
+			}
+		}
+	}
+
+	// Wait for the discovery occurrence to enter a terminal state.
```
nenaddedic, line 347: It seems that you can combine this with the previous loop.  After L342 you check the state, and then good terminal => nil, bad terminal => error, else loop on.

pkg/kritis/metadata/containeranalysis/containeranalysis.go
nenaddedic, line 302: It would also be good to add a test for this function.

pkg/kritis/metadata/containeranalysis/containeranalysis.go
```diff
@@ -299,41 +299,33 @@ func (c Client) DeleteOccurrence(ID string) error {
...
-			return fmt.Errorf("timeout while retrieving discovery occurrence")
-		case <-ticker.C:
+	for {
+		// Waiting for discovery occurrence to appear.
+		if discoveryOccurrence == nil {
 			req := &grafeas.ListOccurrencesRequest{
 				Parent: fmt.Sprintf("projects/%s", getProjectFromContainerImage(containerImage)),
 				// Vulnerability discovery occurrences are always associated with the
 				// PACKAGE_VULNERABILITY note in the "goog-analysis" GCP project.
 				Filter: fmt.Sprintf(`resourceUrl=%q AND noteProjectId="goog-analysis" AND noteId="PACKAGE_VULNERABILITY"`, util.GetResourceURL(containerImage)),
```
nenaddedic, line 321: Should this also filter by Note Kind?

pkg/kritis/metadata/containeranalysis/containeranalysis.go
```diff
@@ -299,41 +299,33 @@ func (c Client) DeleteOccurrence(ID string) error {
 
 // Poll discovery occurrence for an image and wait until container analysis
 // finishes. Throws an error if analysis is not successful or timeouts.
-func WaitForVulnzAnalysis(containerImage string, timeout time.Duration) error {
+func (c Client) WaitForVulnzAnalysis(containerImage string, timeout time.Duration) error {
```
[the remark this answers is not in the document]
nenaddedic: OK, I see your previous response, I think you're basically saying that vulnz scanning isn't a grafeas feature.  But as long as the Note and Occurrence of the appropriate Kind exist, this method should still work, correct?

pkg/kritis/metadata/containeranalysis/containeranalysis.go
ooq: noteProjectId + noteId should be sufficient for the unique note.

pkg/kritis/metadata/containeranalysis/containeranalysis.go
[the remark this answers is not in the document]
ooq: I added methods in `metadata.go` and all implementations of `ReadWriteClient`.

ooq:
Hi @nenaddedic , addressed comments.
The only thing is that no tests were added for the method. Although `WaitForVulnzAnalysis` is also added to `MockMetadataClient`, no methods other than signer main was calling `WaitForVulnzAnalysis`.

integration/signer/tests/test-bypass-and-sign.sh
```diff
@@ -0,0 +1,49 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+echo ""
+echo ""
+
+set -ex
```
nenaddedic, line 20: set -eux perhaps?
to error on unset vars, such as $PROJECT_ID

pkg/kritis/metadata/grafeas/grafeas.go
```diff
@@ -274,3 +275,76 @@ func (c Client) fetchAttestationOccurrence(containerImage string, kind string, a
...
+	tok := strings.Split(image, "/")
+	if len(tok) < 2 {
+		return ""
+	}
+	return tok[1]
+}
+
+// Poll discovery occurrence for an image and wait until container analysis
+// finishes. Throws an error if analysis is not successful or timeouts.
+func (c Client) WaitForVulnzAnalysis(containerImage string, timeout time.Duration) error {
```
nenaddedic, line 289: Is this code any different than the corresponding code in containeranalysis.go?  Is it possible and useful to share most of it?

pkg/kritis/metadata/grafeas/grafeas.go
```diff
@@ -274,3 +275,76 @@ func (c Client) fetchAttestationOccurrence(containerImage string, kind string, a
 	}
 	return occs, nil
 }
+
+func getProjectFromContainerImage(image string) string {
```
nenaddedic, line 279: It seems like this already exists in the containeranalysis client, no?  Can we move it to some common place?

integration/signer/tests/test-bypass-and-sign.sh
ooq: ack.

pkg/kritis/metadata/grafeas/grafeas.go
ooq: The main reason for code reuse here is different client libraries used:`cloud.google.com/go/containeranalysis/apiv1beta1 ` vs. `google.golang.org/genproto/googleapis/devtools/containeranalysis/v1beta1`, then if we want to put this in a common function, that function needs to create a common layer over the two clients in some way, which I think it's a bit complicated at this point.

TBH, I don't know why they need to use different client libraries to begin with, which is an investigation not merited at the moment IMO.

pkg/kritis/metadata/grafeas/grafeas.go
ooq: Ack. Yeah, we should reuse this.

PR #528, opened 2020-06-05
Remove need of service account credentials for signer

ooq:
- When running in local environment, a user can set credentials with environment variable `GOOGLE_APPLICATION_CREDENTIALS`.
- In Cloud Build, this is automatically set.


ooq:
Travis test passed, no idea why it's not looping out the results.

nenaddedic:
IIUC this does not prevent the user to use an exported key, it just changes how to do that if that's what they want (env var vs flag).

ooq:
@nenaddedic this actually does remove the flag option. 
We could add it back? WDYT?
Merging this for now.

PR #529, opened 2020-06-09
[No review] [no merge] PR to test if Travis still works properly.

ooq:


ooq:
Summary: we observed that Travis status was not updated properly on the PR. After multiple tries with this PR, it looks like the problem was temporary.

Issue #530, opened 2020-06-10
Add Key ID check in cryptolib's NewPublicKey


acamadeo:
Validate the user-provided key ID in [NewPublicKey](https://github.com/grafeas/kritis/blob/d88b662d2c1797871b32c58e65b702fff35ad602/pkg/kritis/cryptolib/verifier.go#L53). For PKIX and JWT, keys ID's should only contain valid URI characters, but does need to be a valid URI. For PGP, the key ID should match the key's OpenPGP RFC4880 V4 fingerprint. Return an error if conditions are not satisfied.

PR #532, opened 2020-06-10
Validate public key ID in NewPublicKey

acamadeo:
PGP key ID's should match key fingerprint. PKIX and JWT key ID's should contain valid URI characters.

Bug: #530


pkg/kritis/cryptolib/verifier.go
```diff
@@ -49,13 +52,47 @@ type PublicKey struct {
...
+func NewPublicKey(keyType KeyType, keyData []byte, keyID string) (*PublicKey, error) {
+	switch keyType {
+	case Pgp:
+		err := validatePgpKeyID(keyData, keyID)
+		if err != nil {
+			return nil, err
+		}
+	case Pkix, Jwt:
+		// Valid URI characters (see http://tools.ietf.org/html/rfc3986#section-2)
+		reURI := regexp.MustCompile(`^[a-zA-Z0-9-._~:\/?#\[\]@!$&'\(\)*+,;=%]+$`)
```
nenaddedic, line 66: Is there any way to use url.Parse() for validation instead of this hard to understand regexp?

acamadeo: Patrick advised me not to parse the key ID as a URI, but instead to check that it contains valid URI characters. He said that strong enforcement of URI-compatibility is deprecated in Binauthz.

See comment: https://docs.google.com/document/d/1u6V2pPrIrA4xFR3J2pcqjZajkuT5dZ8_Cck2co1prfg/edit?disco=AAAAGbNnOaM

pkg/kritis/cryptolib/verifier_test.go
```diff
@@ -24,15 +24,123 @@ import (
...
+			name:        "unknown key type",
+			keyType:     UnknownKeyType,
+			expectedErr: true,
+		},
+	}
+	for _, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			_, err := NewPublicKey(tc.keyType, tc.keyData, tc.keyID)
+			if tc.expectedErr && err == nil {
+				t.Errorf("Expected error, but returned none")
```
alexcope, line 118: In go test error messages, put got before want, e.x. "got nil err, expected not-nil"

pkg/kritis/cryptolib/verifier.go
```diff
@@ -49,13 +52,47 @@ type PublicKey struct {
...
+	keyring, err := openpgp.ReadArmoredKeyRing(bytes.NewReader(keyData))
+	if err != nil {
+		return fmt.Errorf("error reading armored public key: %v", err)
+	}
+	if len(keyring) != 1 {
+		return fmt.Errorf("expected 1 public key, got %d", len(keyring))
+	}
+	key := keyring[0]
+	if keyID != fmt.Sprintf("%X", key.PrimaryKey.Fingerprint) {
+		return fmt.Errorf("keyID does not match key fingerprint")
```
alexcope, line 93: nit: include the kid and fingerprint in the error message

pkg/kritis/review/validating_transport.go
```diff
@@ -126,8 +126,12 @@ func (avt *AttestorValidatingTransport) parsePublicKeys(attestorKeys []v1beta1.P
 			invalidKeys = append(invalidKeys, attestorKey.KeyId)
 			continue
 		}
-		publicKey := cryptolib.NewPublicKey(cryptolib.Pgp, decodedKey, attestorKey.KeyId)
-		publicKeys = append(publicKeys, publicKey)
+		publicKey, err := cryptolib.NewPublicKey(cryptolib.Pgp, decodedKey, attestorKey.KeyId)
+		if err != nil {
+			glog.Warningf("Error creating PublicKey: %v", err)
```
alexcope, line 131: invalidKeys = append(invalidKeys, attestorKey.KeyId)?

pkg/kritis/cryptolib/verifier.go
```diff
@@ -49,13 +52,47 @@ type PublicKey struct {
...
+	case Pgp:
+		err := validatePgpKeyID(keyData, keyID)
+		if err != nil {
+			return nil, err
+		}
+	case Pkix, Jwt:
+		// Valid URI characters (see http://tools.ietf.org/html/rfc3986#section-2)
+		reURI := regexp.MustCompile(`^[a-zA-Z0-9-._~:\/?#\[\]@!$&'\(\)*+,;=%]+$`)
+		if !reURI.MatchString(keyID) {
+			return nil, fmt.Errorf("key ID contains invalid characters")
```
alexcope, line 68: It is very difficult for any user to debug this issue without digging into the code. This error message should have some way for users to figure out the expected format.

pkg/kritis/cryptolib/verifier.go
alexcope, line 66: This seems like a reasonable restriction, but make sure it's documented in the public protos.
From discussions with Patrick, we have been thinking of Key Ids as a StringOrUri (https://tools.ietf.org/html/rfc7519) but I don't think we have formally defined them as such.

pkg/kritis/cryptolib/verifier.go
```diff
@@ -49,13 +52,47 @@ type PublicKey struct {
...
-func NewPublicKey(keyType KeyType, keyData []byte, keyID string) PublicKey {
-	return PublicKey{
+// this should be the OpenPGP RFC4880 V4 fingerprint of the key. For PKIX and
+// JWT, the ID should contain valid URI characters.
+func NewPublicKey(keyType KeyType, keyData []byte, keyID string) (*PublicKey, error) {
+	switch keyType {
+	case Pgp:
+		err := validatePgpKeyID(keyData, keyID)
+		if err != nil {
+			return nil, err
```
ooq, line 62: In BinAuthz public doc: https://cloud.google.com/binary-authorization/docs/reference/rest/v1/projects.attestors#attestorpublickey
PGP key id would be overwritten by API generated ID. Should we follow this behavior instead of throwing an error?

pkg/kritis/cryptolib/verifier.go
ooq, line 68: Same as above, should we generate a ID for user if it's not provided for pkix case? See: https://cloud.google.com/binary-authorization/docs/reference/rest/v1/projects.attestors#attestorpublickey

pkg/kritis/cryptolib/verifier_test.go
```diff
@@ -24,15 +24,123 @@ import (
...
+		{
+			name:        "valid PKIX key ID",
+			keyType:     Pkix,
+			keyID:       "valid-key-id",
+			expectedErr: false,
+		},
+		{
+			name:        "invalid PKIX key ID",
+			keyType:     Pkix,
+			keyID:       "^{invalid-key-id}",
```
ooq, line 105: Just curious, why `^{invalid-key-id}`?

pkg/kritis/cryptolib/verifier_test.go
acamadeo: Done.

pkg/kritis/cryptolib/verifier.go

pkg/kritis/review/validating_transport.go
acamadeo: Done.

pkg/kritis/cryptolib/verifier.go

pkg/kritis/cryptolib/verifier.go
acamadeo: Switched to StringOrUri.

pkg/kritis/cryptolib/verifier.go
acamadeo: Done. Justification:

In Binauthz, the ID is overwritten during the creation of the `Attestor` resource. In Kritis, it would be ideal to replicate this behavior at `AttestationAuthority` creation time, but we do not have a management service.

However, by enforcing this behavior in NewPublicKey(), we can minimize false negative errors in VerifyAttestation(). For PGP, the ID of an attestation comes directly from the private key (see [here](https://github.com/grafeas/kritis/blob/257ac88f6544df5a89ae4a600811f41eb90dcec5/pkg/kritis/cryptolib/pgp.go#L87)). Introducing this behavior allows users to use public keys which can correctly verify attestations, even if its user-provided ID is incorrect.

Note: for PKIX/JWT, we will need the Signer implementations to conform the attestation ID as it is done here.

pkg/kritis/cryptolib/verifier.go
acamadeo: See above.

pkg/kritis/cryptolib/verifier_test.go
acamadeo: Replaced with an invalid URI.

pkg/kritis/cryptolib/verifier.go
```diff
@@ -42,20 +46,67 @@ type PublicKey struct {
...
+		}
+		newKeyID = id
+	case Pkix, Jwt:
+		id, err := extractPkixKeyID(keyData, keyID)
+		if err != nil {
+			return nil, err
+		}
+		newKeyID = id
+	case UnknownKeyType:
```
ooq, line 76: NIT: this case is not need?

pkg/kritis/cryptolib/verifier.go
```diff
@@ -42,20 +46,67 @@ type PublicKey struct {
...
+	return fmt.Sprintf("%X", keyring[0].PrimaryKey.Fingerprint), nil
+}
+
+func extractPkixKeyID(keyData []byte, keyID string) (string, error) {
+	// TODO(https://github.com/grafeas/kritis/issues/541): Generate ID based on
+	// DER encoding of key when keyID is empty string.
+	if strings.Contains(keyID, ":") {
+		_, err := url.ParseRequestURI(keyID)
+		if err != nil {
+			return "", fmt.Errorf("keyID not formatted as StringOrURI: must either not contain \":\" or be valid URI")
```
ooq, line 106: Include `keyID` as part of error message.

pkg/kritis/cryptolib/verifier.go
```diff
@@ -42,20 +46,67 @@ type PublicKey struct {
...
 // public key, one of Pgp, Pkix or Jwt. `keyData` contains the raw key
 // material. `keyID` contains a unique identifier for the public key. For PGP,
-// this should be the OpenPGP RFC4880 V4 fingerprint of the key.
-func NewPublicKey(keyType KeyType, keyData []byte, keyID string) PublicKey {
-	return PublicKey{
+// this field should be left blank. The ID will be the OpenPGP RFC4880 V4
+// fingerprint of the key. For PKIX and JWT, this may be left blank, and the ID
+// will be generated based on the DER encoding of the key. If not blank, the ID
+// should be a StringOrURI: it must either not contain ":" or be a valid URI.
+func NewPublicKey(keyType KeyType, keyData []byte, keyID string) (*PublicKey, error) {
```
ooq, line 61: Just curious, why return `*PublicKey` instead of `PublicKey`?

pkg/kritis/cryptolib/verifier.go
acamadeo: Done, removed

pkg/kritis/cryptolib/verifier.go

pkg/kritis/cryptolib/verifier.go
acamadeo: Following Go convention where constructor named `NewThing()` returns `*Thing`.

ooq: Gotcha. Thanks!

acamadeo:

PR #533, opened 2020-06-12
Integrate Cryptolib Signer into Kritis

acamadeo:
Replace `util.CreateAttestationSignature` with `util.CreateAttestation`, which uses `cryptolib.Signer` to create an attestation for an image. 

Notes: 
- Base64 regression testcases have been deleted, as it is assumed that `cryptolib` does not base64-encode signatures.
- `createEntityFromKeys` is copied from [attestation.go](https://github.com/grafeas/kritis/blob/537f5e9f2fae78ffe875091ff0a1a01adc288c6a/pkg/kritis/attestation/attestation.go#L81), which will be deprecated.


pkg/kritis/review/review_test.go
```diff
@@ -308,31 +289,23 @@ func TestReviewISP(t *testing.T) {
...
 	}
 
 	noVulnImage := testutil.IntTestImage
-	sigNoVuln, err := util.CreateAttestationSignature(noVulnImage, sec)
+	attNoVuln, err := util.CreateAttestation(noVulnImage, sec)
 	if err != nil {
 		t.Fatalf("unexpected error %v", err)
 	}
+
+	validAtts := []cryptolib.Attestation{*attVuln}
```
alexcope, line 303: This could use a comment. The attestation on a vulnerable image the valid one?

pkg/kritis/util/util.go
```diff
@@ -36,16 +60,94 @@ func GetResource(image string) *grafeas.Resource {
...
+	armorWriter.Close()
+	return keyBuffer.Bytes(), nil
+}
+
+func createEntityFromKeys(pubKey *packet.PublicKey, privKey *packet.PrivateKey) (*openpgp.Entity, error) {
+	currentTime := pgpConfig.Now()
+	uid := packet.NewUserId("", "", "")
+	if uid == nil {
+		return nil, errors.New("user id field contained invalid characters")
+	}
```
alexcope, line 112: This error does not make sense because NewUserId is taking no arguments

pkg/kritis/util/util.go
```diff
@@ -36,16 +60,94 @@ func GetResource(image string) *grafeas.Resource {
 	return &grafeas.Resource{Uri: GetResourceURL(image)}
 }
 
-func CreateAttestationSignature(image string, pgpSigningKey *secrets.PGPSigningSecret) (string, error) {
-	hostSig, err := container.NewAtomicContainerSig(image, map[string]string{})
+func CreateAttestation(image string, pgpSigningKey *secrets.PGPSigningSecret) (*cryptolib.Attestation, error) {
```
ooq, line 63: `PGPSigningSecret` contains public key information. Can we refactor it to remove public key? (Could be done in a different PR).

pkg/kritis/util/util.go
```diff
@@ -36,16 +60,94 @@ func GetResource(image string) *grafeas.Resource {
...
+	}
+
+	att, err := signer.CreateAttestation([]byte(payload))
+	if err != nil {
+		return nil, err
+	}
+	return att, nil
+}
+
+func extractPrivateKeyBytes(secret *secrets.PGPSigningSecret) ([]byte, error) {
```
ooq, line 89: Question/investigation note: is there an easier way to get private key from secret? Maybe this is something we should factor into secret package?

ooq: Discussed with @acamadeo. There is a concern of maintenance over leaving the these crypto code in `util` package. We'll consider refactor `CreateAttestation` interface or `secrets.PGPSigningSecret` to simply code here.

docs/standalone/create_attestation.go
```diff
@@ -128,21 +128,19 @@ func createOccRequest(note *grafeas.Note) *grafeas.CreateOccurrenceRequest {
...
 
 	// Create Attestation Signature
-	sig, err := util.CreateAttestationSignature(Image, s)
+	att, err := util.CreateAttestation(Image, s)
 	if err != nil {
 		log.Fatal(err)
 		return nil
 	}
-	log.Printf("Created attestation signature: %v", sig)
+	log.Printf("Created attestation signature: %v", string(att.Signature))
```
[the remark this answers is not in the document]
acamadeo: Done.

pkg/kritis/review/review_test.go
acamadeo: Done. Renamed and added comment

pkg/kritis/util/util.go
acamadeo: Done.

pkg/kritis/util/util.go
acamadeo: Will refactor in another PR

pkg/kritis/util/util.go
acamadeo: After some research, it is not possible to extract a usable private key payload from a `packet.PrivateKey`. You can extract one from an `openpgp.Entity`. 

Currently, `secrets.PGPSigningSecret` holds a `packet.PublicKey` and `packet.PrivateKey`. I will write a PR that refactors `secrets.PGPSigningSecret` to instead hold an `openpgp.Entity` which contains the public and private key.

acamadeo:

PR #534, opened 2020-06-12
Delete unused attestation file

acamadeo:
All the functionality from [attestation.go](https://github.com/grafeas/kritis/blob/537f5e9f2fae78ffe875091ff0a1a01adc288c6a/pkg/kritis/attestation/attestation.go#L81) has either moved to other files or is unused.


acamadeo:

PR #538, opened 2020-06-17
Add bypass-and-sign mode in signer

ooq:
`bypass-and-sign` mode will create attestation for an image without any policy check.

### DIFF
--- a/cmd/kritis/signer/main.go
+++ b/cmd/kritis/signer/main.go
@@ -119,6 +119,10 @@ func main() {
 		Project:   policy.Spec.Project,
 	})
 
+	if image == "" {
+		glog.Fatalf("Image url is empty: %s", image)
+	}
+	
 	if SignerMode(mode) == BypassAndSign {
 		r.SignImage(image)
 		return
@@ -126,10 +130,6 @@ func main() {
 
 	if SignerMode(mode) == CheckAndSign {
 		// Read the vulnz scanning events
-		if image == "" {
-			glog.Fatalf("Image url is empty: %s", image)
-		}
-
 		vulnz, err := client.Vulnerabilities(image)
 		if err != nil {
 			glog.Fatalf("Found err %s", err)
--- a/cmd/kritis/signer/main.go
+++ b/cmd/kritis/signer/main.go
@@ -32,9 +32,16 @@ import (
 	yaml "k8s.io/apimachinery/pkg/util/yaml"
 )
 
-func main() {
-	var image, pri_key_path, passphrase, pub_key_path, policy_path string
+const (
+	SignerMode_CheckAndSign  = "check-and-sign"
+	SignerMode_CheckOnly     = "check-only"
+	SignerMode_BypassAndSign = "bypass-and-sign"
+)
 
+func main() {
+	var image, pri_key_path, passphrase, pub_key_path, policy_path, mode string
+
+	flag.StringVar(&mode, "mode", "", "mode of operation, check-and-sign|check-only|bypass-and-sign")
 	flag.StringVar(&image, "image", "", "image url, e.g., gcr.io/foo/bar@sha256:abcd")
 	flag.StringVar(&pri_key_path, "private_key", "", "signer private key path, e.g., /dev/shm/key.pgp")
 	flag.StringVar(&passphrase, "passphrase", "", "passphrase for private key, if any")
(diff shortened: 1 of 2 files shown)


cmd/kritis/signer/main.go
```diff
@@ -32,37 +32,43 @@ import (
 	yaml "k8s.io/apimachinery/pkg/util/yaml"
 )
 
+const (
+	SignerMode_CheckAndSign  = "check-and-sign"
+	SignerMode_CheckOnly     = "check-only"
+	SignerMode_BypassAndSign = "bypass-and-sign"
```
nenaddedic, line 38: optional, mega-nit:
always-sign

cmd/kritis/signer/main.go
```diff
@@ -32,37 +32,43 @@ import (
...
+	flag.StringVar(&mode, "mode", "", "mode of operation, check-and-sign|check-only|bypass-and-sign")
 	flag.StringVar(&image, "image", "", "image url, e.g., gcr.io/foo/bar@sha256:abcd")
 	flag.StringVar(&pri_key_path, "private_key", "", "signer private key path, e.g., /dev/shm/key.pgp")
 	flag.StringVar(&passphrase, "passphrase", "", "passphrase for private key, if any")
 	flag.StringVar(&pub_key_path, "public_key", "", "public key path, e.g., /dev/shm/key.pub")
 	flag.StringVar(&policy_path, "policy", "", "vulnerability signing policy file path, e.g., /tmp/vulnz_signing_policy.yaml")
 	flag.Parse()
 
-	glog.Infof("image: %s, s_key: %s, policy: %s", image, pri_key_path, policy_path)
+	if mode != SignerMode_CheckAndSign && mode != SignerMode_BypassAndSign {
```
nenaddedic, line 52: Can you use switch instead?

cmd/kritis/signer/main.go
```diff
@@ -32,37 +32,43 @@ import (
...
+		if mode == SignerMode_CheckOnly {
+			glog.Fatalf("Mode %s note supported yet.", mode)
+		} else {
+			glog.Fatalf("Unrecognized mode %s.", mode)
+		}
+	}
 
 	signerKey, err := ioutil.ReadFile(pri_key_path)
```
nenaddedic, line 60: This is only needed if signing is requested, so could you make sure we don't try to read the key in check-only?

cmd/kritis/signer/main.go
```diff
@@ -32,37 +32,43 @@ import (
...
 	if err != nil {
 		glog.Fatalf("Fail to read signer key: %v", err)
 	}
 
+	// Parse the vulnz signing policy
```
nenaddedic, line 65: And this is only needed if policy check is requested, so could you make sure we don't try to get the policy in always-sign?

cmd/kritis/signer/main.go
ooq: do we check policy for `always-sign`?

cmd/kritis/signer/main.go
ooq: ack.

cmd/kritis/signer/main.go
ooq: Will do this when implementing `check-only` mode, which is not being supported as part of this PR.

cmd/kritis/signer/main.go
ooq: So it is needed now because the policy carries `noteReference`. I think in upcoming PRs, we should remove those config info, i.e., `noteReference` and attestation `project` out of the policy. And the policy should only contain vulnz requirements. WDYT?

cmd/kritis/signer/main.go
nenaddedic: I would say no, it seems odd to check policy, see a failure, and sign anyway.

cmd/kritis/signer/main.go
nenaddedic: Yes, I agree.

cmd/kritis/signer/main.go
ooq: It feels to me `always-sign` implies, or at least does not exclude check. `just-sign`?

cmd/kritis/signer/main.go
```diff
@@ -116,12 +119,12 @@ func main() {
...
-	if mode == SignerMode_BypassAndSign {
+	if SignerMode(mode) == BypassAndSign {
 		r.SignImage(image)
 		return
 	}
 
-	if mode == SignerMode_CheckAndSign {
+	if SignerMode(mode) == CheckAndSign {
 		// Read the vulnz scanning events
 		if image == "" {
```
nenaddedic, line 129: It seems like this error is possible if BypassAndSign too.

pkg/kritis/signer/signer.go
```diff
@@ -75,13 +75,19 @@ func (s Signer) ValidateAndSign(imageVulnz ImageVulnerabilities, vps v1beta1.Vul
...
 
 	glog.Infof("Image %q passes VulnzSigningPolicy %s.", imageVulnz.ImageRef, vps.Name)
-	existed, _ := s.isAttestationAlreadyExist(imageVulnz.ImageRef)
+	return s.SignImage(imageVulnz.ImageRef)
+}
+
+// ValidateAndSign signs an image without doing any policy check.
+// Returns an error if image does not pass or creating an attestation fails.
+func (s Signer) SignImage(image string) error {
+	existed, _ := s.isAttestationAlreadyExist(image)
```
nenaddedic, line 84: Is it safe to ignore this error?  Shouldn't block this PR, but let's think about it.

PR #539, opened 2020-06-17
Create a cryptolib.Signer interface to Cloud KMS

nenaddedic:
I had to update vendor/ because the existing kms client libs didn't have any of the asymmetric crypto methods, which are required for signing.


ooq:
FYI, #518 is merged. Updating with master should pass this PR on the GCB test.
Let's add an integration test for KMS signing as well. As part of it, we probably need to add KMS access to GCB service account.

pkg/kritis/signer/cloud_kms.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (s kmsSigner) CreateAttestation(payload []byte) (*cryptolib.Attestation, error) {
+	var digest hash.Hash
+	switch s.digestAlg {
+	case SHA256:
+		digest = sha256.New()
+	case SHA384:
+		digest = sha512.New384()
+	case SHA512:
+		digest = sha512.New()
```
ooq, line 67: Error handling for non-supported cases?

hack/dep.sh
```diff
@@ -53,11 +53,8 @@ files=( $(validate_diff --name-only -- 'Gopkg.toml' 'Gopkg.lock' 'vendor/' || tr
 unset IFS
 
 if [ ${#files[@]} -gt 0 ]; then
-	if ! [ -x "$(command -v dep)" ]; then
-		install_dep
```
ooq, line 57: This shouldn't be removed?

nenaddedic: Oops, done.

pkg/kritis/signer/cloud_kms.go
nenaddedic: Done.

PR #542, opened 2020-06-22
Update vulnerability signing policy spec

ooq:
This does update the `vulnzsigningpolicy` spec:

s/`packageVulnerabilityRequirements`/`imageVulnerabilityRequirements`/g
s/`maximumSeverity`/`maximumFixableSeverity`/g
s/`maximumFixNotAvailableSeverity`/`maximumUnfixableSeverity`/g


ooq:
@nenaddedic Done.

pkg/kritis/crd/vulnzsigningpolicy/violation.go
```diff
@@ -63,24 +63,24 @@ func UnqualifiedImageReason(image string) policy.Reason {
 	return policy.Reason(fmt.Sprintf("%s is not a fully qualified image. You can run 'kubectl plugin resolve-tags' to qualify all images with a digest.", image))
 }
 
-// FixUnavailabileReason returns a detailed reason if an unfixable CVE exceeds max severity
-func FixUnavailableReason(image string, v metadata.Vulnerability, vsp v1beta1.VulnzSigningPolicy) policy.Reason {
-	ms := vsp.Spec.PackageVulnerabilityRequirements.MaximumFixUnavailableSeverity
+// UnfixableSeverityViolationReason returns a detailed reason if an unfixable CVE exceeds max severity
```
nenaddedic, line 66: Does this file need to change?  Isn't this only used in kritis server (not in the signer)?

ooq: This is used as part of signer's logic to find violations during policy check. This file is not used in kritis server just as any part of `vulnzsigningpolicy`.

nenaddedic: Oh, I think I see what is going on.  We're following a pattern used in Kritis server, which is probably not appropriate here in the signer.  ValidateFunc spits out Violations, which can then be Handle()d per Strategy.  I think we don't need most of that.  Nothing to do with this PR though, let's discuss separately.

PR #543, opened 2020-06-22
Remove attestation project and note name from policy spec

ooq:


ooq:
Travis test passed but is experiencing delayed report-back. 
Kokoro failures are flaky.
Merging this.

PR #544, opened 2020-06-22
Support check only mode

ooq:


pkg/kritis/signer/signer.go
```diff
@@ -61,24 +57,7 @@ var (
...
-		return fmt.Errorf("image %q does not pass VulnzSigningPolicy %q: %v", imageVulnz.ImageRef, vps.Name, violations)
-	}
-
-	glog.Infof("Image %q passes VulnzSigningPolicy %s.", imageVulnz.ImageRef, vps.Name)
-	return s.SignImage(imageVulnz.ImageRef)
-}
-
-// ValidateAndSign signs an image without doing any policy check.
+// SignImage signs an image without doing any policy check.
 // Returns an error if image does not pass or creating an attestation fails.
```
nenaddedic, line 61: Update comment: there's nothing to pass.

ooq: Done.

PR #545, opened 2020-06-23
Reduce demo scope

donmccasland:
Reducing scope of demo to just policy check.  Kritis-signer now polls container analysis until results are available.


ooq:
Thanks for fixing the demo @donmccasland !
I also opened a PR (#547) to update some other parts of demo code. PTAL

samples/signer/setup.sh
```diff
@@ -80,99 +77,13 @@ binauthz_project_setup () {
...
-    defaultAdmissionRule:
-      evaluationMode: REQUIRE_ATTESTATION
-      enforcementMode: ENFORCED_BLOCK_AND_AUDIT_LOG
-      requireAttestationsBy:
-        - projects/${BINAUTHZ_PROJECT}/attestors/${ATTESTOR}
-    name: projects/${BINAUTHZ_PROJECT}/policy
-EOM
-  gcloud container binauthz policy import ${DIR}/binauthz-policy.yaml
-
-  cat policy_template.yaml \
```
ooq, line 168: We would probably still need `policy.yaml`. I should be good enough to do `cp policy_template.yaml policy.yaml` once #547 is merged.

PR #547, opened 2020-06-24
Update command line parameters and policy in signer demo

ooq:
In conjunction with #545, this PR should reduce the signer demo's scope to policy check.
It also updates the policy template so that it conforms to new policy spec.
It then adds a new allowlisted CVE. (The Debian10 image was discovered a new vulnz that broke the "good" case. The added allowlisted CVE will fix that.)


ooq:
cc @donmccasland

PR #550, opened 2020-06-24
Move PublicKey into its own file

acamadeo:
Minor refactor: separate `PublicKey` definition and constructor logic from `Verifier` definition and implementation.


acamadeo:

PR #551, opened 2020-06-25
Pinpoint debian versions in Dockerfile

ooq:
Use digest instead of `latest` tag in those Dockerfile, so new Debian releases would not change vulnerability results completely.

However, vulnerability results could still change for the images, as new vulnerabilities are being discovered.

This should make signer integration testing and samples more stable.


ooq:
cc @donmccasland

PR #552, opened 2020-06-25
Adding and correcting cloud build permissions

donmccasland:


samples/signer/setup.sh
```diff
@@ -58,21 +58,12 @@ binauthz_project_setup () {
   set +x; echo "Setting up service accounts and permissions.."; set -x
   gcloud projects add-iam-policy-binding ${BINAUTHZ_PROJECT} \
     --member serviceAccount:${BINAUTHZ_PROJECTNUM}@cloudbuild.gserviceaccount.com \
-    --role roles/container.developer
+    --role roles/containeranalysis.notes.occurrences.viewer
 
-  gcloud iam service-accounts create kritis-signer \
-    --description "For creating attestations" \
-    --display-name "kritis-signer"
   gcloud projects add-iam-policy-binding ${BINAUTHZ_PROJECT} \
```
ooq, line 63: Because the `check-only` mode does not create attestations, it also does not require KMS.

PR #553, opened 2020-06-25
Update setup.sh

siliconmoo:
Update comment section to remove references to signing and uploading and attestation?


ooq:
Thanks for the fix. Merged.

PR #554, opened 2020-06-26
Update new public key to use signatureAlgorithm field.

treyridley:
Have the NewPublicKey create a public key with the signatureAlgorithm field populated. This is needed for JWT and PKIX verification. For PGP, the signatureAlgorithm should be undefined.


pkg/kritis/cryptolib/public_key.go
```diff
@@ -55,12 +55,18 @@ func NewPublicKey(keyType KeyType, keyData []byte, keyID string) (*PublicKey, er
 			return nil, err
 		}
 		newKeyID = id
+		if signatureAlgorithm != UndefinedSigningAlgorithm {
```
alexcope, line 58: For code readability reasons, I'm hesitant to say that all pgp keys have undefined signing algorithm going forward. 

If we need this for backwards compatibility, maybe we can support both undefined and a new "pgp" signature algorithm. Someone who is more knowledgeable about kirtis should weigh in.

acamadeo: I think it would make sense to let PGP keys have a new PGP SignatureAlgorithm. 

Perhaps it makes sense to split `NewPublicKey` by KeyType? You could do the following:
```
NewPgpPublicKey(keyData []byte, keyID string) (*PublicKey, error)
NewPkixPublicKey(signatureAlgorithm SignatureAlgorithm, keyData []byte, keyID string) (*PublicKey, error)
NewJwtPublicKey(signatureAlgorithm SignatureAlgorithm, keyData []byte, keyID string) (*PublicKey, error)
```

pkg/kritis/cryptolib/verifier_test.go
```diff
@@ -73,51 +73,72 @@ const verifierPublicKeyID = "446625EB36036D7546B2D63B77B4BE6989834233"
...
+			keyType:            Pkix,
+			signatureAlgorithm: UndefinedSigningAlgorithm,
+			keyID:              "valid-key-id",
+			expectedErr:        true,
 		},
 		{
-			name:        "unknown key type",
-			keyType:     UnknownKeyType,
-			expectedErr: true,
+			name:               "valid PGP key ID with defined signature algorithm",
```
acamadeo, line 118: nit: "PGP key with incorrect signature algorithm"

pkg/kritis/cryptolib/verifier_test.go
```diff
@@ -73,51 +73,72 @@ const verifierPublicKeyID = "446625EB36036D7546B2D63B77B4BE6989834233"
...
+			keyID:              "valid-key-id",
+			expectedErr:        false,
+			expectedID:         "valid-key-id",
 		},
 		{
-			name:        "invalid PKIX key ID",
-			keyType:     Pkix,
-			keyID:       ":{invalid-key-id}",
-			expectedErr: true,
+			name:               "valid PKIX key ID with undefined signature algorithm",
```
acamadeo, line 111: nit: make testcase name highlight the error: "PKIX key with incorrect signature algorithm"

pkg/kritis/cryptolib/public_key.go
treyridley: If we add the PGP Signature algorithm there would be no need to split by KeyType right? Also, the idea behind using it as undefined was to highlight the fact that it was not used in this case. Would something like UnusedPGP work to ensure that it is understood that this is simply a placeholder?

Another question would be are there future plans that would make this field necessary in the PGP case?

alexcope: As we discussed in standup a while back, for now just change this to UnusedPGP or PGPSigningAlgorithmNotApplicable

ooq: Ditto to @alexcope . And I also think makes sense to split the constructors by key type, but it doesn't have to be done in this PR.

treyridley: done.

PR #557, opened 2020-06-30
Add UploadAttestationOccurrence() method in metadata client interface

ooq:
Add `UploadAttestationOccurrence()` method in metadata client interface and all its implementations.

A helper method `CreateOccurrenceFromAttestation()` and its tests are added in the process to deduplicate code.

The existing `CreateAttestationOccurrence()` will now use `UploadAttestationOccurrence()` to upload attestation after creating it. In the long term, we consider deleting `CreateAttestationOccurrence()` altogether because attestation creation should be the caller's responsibility instead of the metadata client. 

`UploadAttestationOccurrence()` also supports both `GenericSignedAttestation` and `PgpSignedAttestation`, while `CreateAttestationOccurrence()` only supports the latter.

Testing for `UploadAttestationOccurrence()` is covered by existing `CreateAttestationOccurrence()` tests, because `CreateAttestationOccurrence()` is the only caller of `UploadAttestationOccurrence()` as for now.


pkg/kritis/metadata/metadata.go
```diff
@@ -144,3 +153,59 @@ func GetAttestationsFromOccurrence(occ *grafeas.Occurrence) ([]cryptolib.Attesta
 	}
 	return atts, nil
 }
+
+// CreateOccurrenceFromAttestation creates an occurrence from an attestation by specified signature type.
+// The created occurrence can either be a PgpSignedAttestation occurrence or a GenericSignedAttestation occurrence.
+func CreateOccurrenceFromAttestation(att *cryptolib.Attestation, containerImage string, noteName string, sType SignatureType) (*grafeas.Occurrence, error) {
+	attestation := &attestationpb.Attestation{}
+
```
nenaddedic, line 161: mega-nit: remove newline

pkg/kritis/metadata/metadata.go
```diff
@@ -144,3 +153,59 @@ func GetAttestationsFromOccurrence(occ *grafeas.Occurrence) ([]cryptolib.Attesta
 	}
 	return atts, nil
 }
+
+// CreateOccurrenceFromAttestation creates an occurrence from an attestation by specified signature type.
+// The created occurrence can either be a PgpSignedAttestation occurrence or a GenericSignedAttestation occurrence.
+func CreateOccurrenceFromAttestation(att *cryptolib.Attestation, containerImage string, noteName string, sType SignatureType) (*grafeas.Occurrence, error) {
+	attestation := &attestationpb.Attestation{}
```
nenaddedic, line 160: nil ?

pkg/kritis/metadata/metadata.go
```diff
@@ -144,3 +153,59 @@ func GetAttestationsFromOccurrence(occ *grafeas.Occurrence) ([]cryptolib.Attesta
...
+		attestation = &attestationpb.Attestation{
+			Signature: &attestationpb.Attestation_GenericSignedAttestation{
+				GenericSignedAttestation: genericSignedAttestation,
+			},
+		}
+	default:
+		return nil, fmt.Errorf("unknown signature type %v", sType)
+	}
+
+	attestationDetails := &grafeas.Occurrence_Attestation{
```
nenaddedic, line 199: NIt: Can you inline this below?

pkg/kritis/metadata/metadata.go
```diff
@@ -144,3 +153,59 @@ func GetAttestationsFromOccurrence(occ *grafeas.Occurrence) ([]cryptolib.Attesta
 	}
 	return atts, nil
 }
+
+// CreateOccurrenceFromAttestation creates an occurrence from an attestation by specified signature type.
+// The created occurrence can either be a PgpSignedAttestation occurrence or a GenericSignedAttestation occurrence.
+func CreateOccurrenceFromAttestation(att *cryptolib.Attestation, containerImage string, noteName string, sType SignatureType) (*grafeas.Occurrence, error) {
```
nenaddedic, line 159: Do we want this to be an exported function?

pkg/kritis/metadata/metadata.go
ooq: Done.

pkg/kritis/metadata/metadata.go
ooq: Done.

pkg/kritis/metadata/metadata.go
ooq: nil doesn't seem to make compiler happy :(

pkg/kritis/metadata/metadata.go
ooq: It needs to be exported to be used in `containeranalysis.go`, etc.

PR #561, opened 2020-07-06
Rename KeyType to AuthenticatorType

acamadeo:
`KeyType` is an enum with values {Pgp, Jwt, Pkix}. This can be confusing, since "Jwt" is not a type of cryptographic key, but rather a signature transport format. `KeyType` is used in cryptolib to indicate the transport format of the Attestation and how to extract meaningful information from it during verification.

### DIFF
--- a/pkg/kritis/cryptolib/public_key.go
+++ b/pkg/kritis/cryptolib/public_key.go
@@ -41,17 +41,17 @@ type PublicKey struct {
 }
 
 // NewPublicKey creates a new PublicKey.
-// `authType` indicates the transport format of the Attestation this PublicKey
-// verifies, one of Pgp, Pkix or Jwt.
+// `authenticatorType` indicates the transport format of the Attestation this
+// PublicKey verifies, one of Pgp, Pkix or Jwt.
 // `keyData` contains the raw key material.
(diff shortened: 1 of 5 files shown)


pkg/kritis/cryptolib/public_key.go
```diff
@@ -39,16 +40,18 @@ type PublicKey struct {
...
+// `keyData` contains the raw key material.
+// `keyID` contains a unique identifier for the public key. For PGP, this field
+// should be left blank. The ID will be the OpenPGP RFC4880 V4 fingerprint of
+// the key. For PKIX and JWT, this may be left blank, and the ID  will be
+// generated based on the DER encoding of the key. If not blank, the ID should
+// be a StringOrURI: it must either not contain ":" or be a valid URI.
+func NewPublicKey(authType AuthenticatorType, keyData []byte, keyID string) (*PublicKey, error) {
 	newKeyID := ""
-	switch keyType {
+	switch authType {
```
alexcope, line 54: nit: authType -> authenticatorType, because auth is generally used to abbreviate authentication, not authenticator

acamadeo: Done.

acamadeo:
Local integration tests passed

PR #562, opened 2020-07-07
Move pkg/kritis/cryptolib to pkg/attestlib.

nenaddedic:


pkg/attestlib/attestation.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package cryptolib
+package attestlib
```
ooq, line 17: I'm for saving a syllable over attestationlib :)

PR #564, opened 2020-07-08
Integrate kms signing in signer

ooq:


cmd/kritis/signer/main.go
```diff
@@ -44,22 +44,28 @@ var (
...
+	kmsDigestAlg string
 )
 
 func init() {
 	flag.StringVar(&mode, "mode", "check-and-sign", "mode of operation, check-and-sign|check-only|bypass-and-sign")
 	flag.StringVar(&image, "image", "", "image url, e.g., gcr.io/foo/bar@sha256:abcd")
 	flag.StringVar(&vulnzTimeout, "vulnz_timeout", "5m", "timeout for polling image vulnerability , e.g., 600s, 5m")
-	flag.StringVar(&priKeyPath, "private_key", "", "signer private key path, e.g., /dev/shm/key.pgp")
-	flag.StringVar(&passphrase, "passphrase", "", "passphrase for private key, if any")
+	flag.StringVar(&pgpPriKeyPath, "pgp_private_key", "", "pgp private signing key path, e.g., /dev/shm/key.pgp")
```
nenaddedic, line 62: Not for this PR, I'm just curious about your opinion.

These flags are getting complicated, do you think it makes sense to have a config file instead?  Maybe using --policy would work, as long as the vulnz policy is clearly separated from key specs etc.

ooq: Yeah, I think adding the option of config file does make sense. The two (flags and config file) can probably co-exist.

ooq:
A note: as part of this PR, we now upload attestations as GenericSignatureAttestations, including with pgp signatures.  @nenaddedic

PR #566, opened 2020-07-09
Supoort -overwrite flag in signer

ooq:
A new bool flag `overwrite` in signer that allows overwriting existing attestation.


pkg/kritis/signer/signer.go
```diff
@@ -78,23 +82,36 @@ var (
 // SignImage signs an image without doing any policy check.
 // Returns an error if creating an attestation fails.
 func (s Signer) SignImage(image string) error {
-	existed, _ := s.isAttestationAlreadyExist(image)
+	existed, err := s.isAttestationAlreadyExist(image)
+	if err != nil {
+		return fmt.Errorf("checking existing attestation status failed: %v", err)
+	}
 	if existed {
```
nenaddedic, line 89: Optional: I suggest restructuring the control flow to reduce nesting, bail out early, and make it easier to follow, maybe like this:
if exists && !overwrite {
  return ...
}
if exists && overwrite {
  delete ...
} else {
  doesn't exist
}
create attestation

integration/signer/tests/test-overwrite.sh
```diff
@@ -0,0 +1,56 @@
...
+docker build --no-cache -t $GOOD_IMAGE_URL -f ./Dockerfile.good .
+
+trap 'delete_image $GOOD_IMAGE_URL'  EXIT
+
+# push good image
+docker push $GOOD_IMAGE_URL
+# get image url with digest format
+GOOD_IMG_DIGEST_URL=$(docker image inspect $GOOD_IMAGE_URL --format '{{index .RepoDigests 0}}')
+
+trap 'delete_occ $GOOD_IMG_DIGEST_URL'  EXIT
```
nenaddedic, line 32: Do traps "stack" like this?  At this point, will both delete_occ and delete_image be executed on EXIT?  Is the order specified?

(This is for my education...)

integration/signer/tests/test-overwrite.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# sign good image in bypass mode
```
nenaddedic, line 34: Nit:
If it's bypassing policy checks, then "good image" is a bit misleading.  There's no good or bad images.

integration/signer/tests/test-overwrite.sh
```diff
@@ -0,0 +1,56 @@
...
+-mode=bypass-and-sign \
+-image=${GOOD_IMG_DIGEST_URL} \
+-pgp_private_key=private.key \
+-policy=policy.yaml \
+-note_name=${NOTE_NAME}
+
+sleep 5
+
+# sign good image in bypass mode
+./signer -v 10 \
```
nenaddedic, line 46: This just checks that the signing doesn't fail, right?  It doesn't actually verify that the attestation was overwritten, correct?

pkg/kritis/signer/signer.go
ooq: Done.

integration/signer/tests/test-overwrite.sh
ooq: Actually your concern is valid. They don't stack...fixing them in #567 instead to reduce merge conflicts.

integration/signer/tests/test-overwrite.sh
ooq: Done.

integration/signer/tests/test-overwrite.sh
ooq: Added check to validate that occurrence id before and after overwrite are different.

ooq:
@nenaddedic Addressed comments. PTAL.

integration/signer/tests/test-overwrite.sh
nenaddedic: Thanks for doing this!  And, sorry, my comment was an honest question, I didn't mean to insist that this needs to be tested so thoroughly.  Either way, this is great now!

PR #567, opened 2020-07-10
Add deployment integration in signer integration testing

ooq:
The added commands make sure that signed images can be deployed to a BinAuthz-enabled test cluster (which requires attestation).


integration/signer/signer_int.sh
```diff
@@ -93,6 +93,8 @@ export -f delete_occ
 #### TEST 1: bypass-and-sign mode ####
 ./tests/test-bypass-and-sign.sh
 
+exit 0
```
nenaddedic, line 96: Huh?

ooq:
@nenaddedic Ready for review.

integration/signer/tests/test-bypass-and-sign-with-kms.sh
```diff
@@ -22,14 +22,16 @@ set -eux
 GOOD_IMAGE_URL=gcr.io/$PROJECT_ID/signer-int-good-image:$BUILD_ID
 docker build --no-cache -t $GOOD_IMAGE_URL -f ./Dockerfile.good .
 
-trap 'delete_image $GOOD_IMAGE_URL'  EXIT
+clean_up() { ARG=$?; delete_image $GOOD_IMAGE_URL; exit $ARG;}
```
nenaddedic, line 25: What does the whole ARG thing do? Does it somehow preserve the exit code? Or am I totally misunderstanding it?

ooq: Yeah, this tries to preserve the exit code.

integration/signer/signer_int.sh
```diff
@@ -93,12 +90,27 @@ delete_occ() {
...
+deploy_image() {
+  IMAGE_URL=$1
+  POD_NAME=$2
+  gcloud config set compute/zone us-central1-c
+  gcloud container clusters get-credentials signer-int-test
+  kubectl run --generator=run-pod/v1 --image=$IMAGE_URL $POD_NAME --command ls
+}
+
+delete_pod() {
+  set +ex
```
nenaddedic, line 104: Shouldn't this be in the beginning of the script, not in a function like this?

ooq: We want the testing to stop whenever a test fails and errors, but the clean-up to always run to completion even some cleaning step fails.

integration/signer/tests/test-bypass-and-sign-with-kms.sh
```diff
@@ -41,5 +43,11 @@ trap 'delete_occ $GOOD_IMG_DIGEST_URL'  EXIT
 -policy=policy.yaml \
 -note_name=${NOTE_NAME}
 
+# deploy to a binauthz-enabled cluster signer-int-test
+clean_up() { ARG=$?; delete_image $GOOD_IMAGE_URL; delete_occ $GOOD_IMG_DIGEST_URL; delete_pod signer-int-test-pod; exit $ARG;}
```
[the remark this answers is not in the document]
ooq: I'm not thinking too much, but it feels that using || could be complicated because you also need to consider clean up for success cases?

Ack on documentation.

PR #576, opened 2020-07-22
Update pkix key parser

treyridley:
Update the private key parser to accept both pkcs1 and pkcs8 formatted private keys.


pkg/attestlib/pkix_key.go
```diff
@@ -33,12 +33,28 @@ func parsePkixPrivateKeyPem(privateKey []byte) (interface{}, error) {
 	if len(rest) != 0 {
 		return nil, errors.New("expected one public key")
 	}
-
-	key, err := x509.ParsePKCS8PrivateKey(der.Bytes)
-	if err != nil {
-		return nil, err
+	switch der.Type {
+	case "RSA PRIVATE KEY":
```
alexcope, line 37: extract these out as constants

treyridley: done.

pkg/attestlib/pkix_key.go
```diff
@@ -27,18 +27,38 @@ import (
...
 	}
-
-	key, err := x509.ParsePKCS8PrivateKey(der.Bytes)
-	if err != nil {
-		return nil, err
+	switch der.Type {
+	case rsaPkcs1Key:
+		key, err := x509.ParsePKCS1PrivateKey(der.Bytes)
+		if err != nil {
+			return nil, err
```
ooq, line 44: NIT: can you wrap the error? e.g., `errors.Wrap(err, fmt.Sprintf("parse %s key failed", der.Type))`. Same for below.

ooq:
Thanks for the PR! Approved.
IDK why the cloud build tests failed. Can you push again to trigger a rerun? (I can manually trigger rerun but I found that manual rerun results would not propagate). 
Ping me if the rerun still fails.

ooq:
@treyridley Can you just try to push some change to retrigger the cloud build test? Might just be something flaky.

pkg/attestlib/pkix_key.go
treyridley: done.

PR #577, opened 2020-07-24
Add sample file for signer user guide

ooq:


ooq:
@nenaddedic They will be referenced from user guide (https://github.com/grafeas/kritis/pull/580) and Binary Authorization documentation.

PR #578, opened 2020-07-27
Fix cloud build signer integration test

ooq:
Explicitly read created attestation before deployment.

I believe the test error is caused by read-after-write consistency in occurrence storage. Sometimes an explicit read could trigger consistency...we never know, but at least the tests pass in this PR.


ooq:
@treyridley This seems to fix the issue. (or it may still be flaky) Can you review it?

PR #587, opened 2020-11-11
feature-586: implement automated check and sign on completion of an analysis of an image in GCR

mvanholsteijn:
This pull request is to draw attention to the implementation of a service, which will subscribe to the container analysis occurrences events. After a scan is completed, it will check the vulnerabilities against the policy and if ok, create an attestation.

I would like to get some feedback on this.

### DIFF
--- a/cmd/kritis/signer/server.go
+++ b/cmd/kritis/signer/server.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 binx.io B.V>
+Copyright 2018 Google LLC
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -13,7 +13,6 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-
 package main
 
 import (
--- a/cmd/kritis/signer/server.go
+++ b/cmd/kritis/signer/server.go
@@ -109,7 +109,6 @@ func containerAnalysisEvent(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-
 	if event.Kind != "DISCOVERY" {
 		response := SignResponse{Status: Failed, Message: "ignoring non-DISCOVERY occurrence"}
 		WriteResponse(w, response, http.StatusOK)
--- a/cmd/kritis/signer/server.go
+++ b/cmd/kritis/signer/server.go
@@ -1,0 +1,236 @@
+/*
+Copyright 2020 binx.io B.V>
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"google.golang.org/genproto/googleapis/devtools/containeranalysis/v1beta1/discovery"
+	grafeaspb "google.golang.org/genproto/googleapis/devtools/containeranalysis/v1beta1/grafeas"
+	"net/http"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/golang/glog"
+)
+
+type EvaluationStatus string
+
+const (
+	Ok        EvaluationStatus = "ok"
+	Signed    EvaluationStatus = "signed"
+	NotSigned EvaluationStatus = "not-signed"
+	Failed    EvaluationStatus = "failed"
+)
+
+type SignRequest struct {
+	Image string
+}
+
+type SignResponse struct {
+	Image      string           `json:"image,omitempty"`
+	Status     EvaluationStatus `json:"status,omitempty"`
+	Message    string           `json:"message,omitempty"`
+	Violations []string         `json:"violations,omitempty"`
+}
+
+// write the sign `response` to `w` with status `code`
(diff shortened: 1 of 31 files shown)


ooq:
Hi @mvanholsteijn , thanks for the PR! I have two questions:
- Is it possible to refactor `cmd/kritis/gcr-signer/main.go` and the Cloud Run configuration so that it leverages the existing signer CLI (i.e., `cmd/kritis/signer/main.go`)? reasoning: gcrsigner is essentially a server version of the existing signer CLI, if we could reuse the signer CLI, we could significantly reduce maintenance overhead down the road.
- Can you provide setup steps without terraform?

mvanholsteijn:
Hi @ooq,

both can be done. Give me a few days to do that.

Cheers,

Mark

mvanholsteijn:
Hi @ooq,

I have refactored the server mode into the kritis-signer; both the code and the documentation.  The cmd/kritis/signer/main.go has quite some changes as a result. 

The installation instructions without terraform were really easy :-)

Cheers,

Mark

cmd/kritis/signer/server.go
```diff
@@ -0,0 +1,228 @@
...
+		response.Violations = violations
+		WriteResponse(w, response, violationStatus)
+		return
+	}
+
+	err = DoSign(request.Image)
+	if err != nil {
+		response.Status = Failed
+		response.Message = err.Error()
+		WriteResponse(w, response, http.StatusInternalServerError)
```
nenaddedic, line 188: Should this really be an internal error?  Also, it seems it should be consistent with the error in L172.

cmd/kritis/signer/server.go
```diff
@@ -0,0 +1,228 @@
...
+	if err != nil {
+		response.Status = Failed
+		response.Message = err.Error()
+		WriteResponse(w, response, http.StatusBadRequest)
+		return
+	}
+	if violations != nil {
+		response.Status = Failed
+		response.Violations = violations
+		WriteResponse(w, response, http.StatusUnprocessableEntity)
```
nenaddedic, line 156: StatusBadRequest, like in above cases?

cmd/kritis/signer/server.go
```diff
@@ -0,0 +1,228 @@
...
+	if status != discovery.Discovered_FINISHED_SUCCESS {
+		http.Error(w, fmt.Sprintf("ignoring DISCOVERY occurrence in status %s", string(status)), http.StatusOK)
+		return
+	}
+	image := occurrence.Resource.Uri
+	if strings.HasPrefix(image, "https://") {
+		image = image[8:]
+	}
+	request := SignRequest{Image: image}
+	doCheckAndSign(w, request, http.StatusOK)
```
nenaddedic, line 129: I think this will swallow any errors.  Technically it'll return them back to containeranalysis, but I think they won't even be logged anywhere.  Would it be better to at least log them here?

I'm aware this wouldn't help all that much with debugging, because the user would still have to sift through the server logs; but maybe it's better than nothing.

cmd/kritis/signer/server.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	err = json.Unmarshal(m.Message.Data, &event)
+	if err != nil {
+		http.Error(w, "could not unmarshal container analysis occurrence event from pubsub message", http.StatusBadRequest)
+		return
+	}
+
+	if event.Kind != "DISCOVERY" {
+		http.Error(w, "ignoring non-DISCOVERY occurrence", http.StatusOK)
```
nenaddedic, line 109: Perhaps downgrade logging level.  This doesn't seem like some exceptional circumstance: AIUI pubsub will send all the Occurrences, not just the DISCOVERY ones.

mvanholsteijn: I did not know that http.Error() prints to stderr too. Fixed it so that all conditions will return an appropriate SignResponse.

cmd/kritis/signer/server.go
mvanholsteijn: I am not quite sure what you are referring to here. When the image does not start with `https://` it will be passed in as is. the DoCheck() function will validate whether it is a proper image reference with digest and return an error if it is not.

cmd/kritis/signer/server.go
mvanholsteijn: StatusBadRequest (400) indicates that the input is incorrect and the caller should fix it.
StatusUnprocessableEntity means the input is correct, but we are  unable to process the contained instructions. 

We could change it to 200 (as we are requested to perform a check, which we did) and have the response checked for success or failure. But then it would be a little bit harder to use this /check call in a program flow (for instance curl).

cmd/kritis/signer/server.go
mvanholsteijn: Line 172 is indeed incorrect. I fixed that. It is an internal server error as we failed to do what we were requested to do and the caller has not way of fixing that. We need to do that.

mvanholsteijn:
@nenaddedic I have updated the PR to the latest master and include your review findings. The number of changed files has been reduced as a result.

ooq:
Hi Mark,
Can we revert to your previous approach of having separate folder `cmd/kritis/gcr-signer/` for the server mode? Let's also maintain the tutorials to be in separate docs. 
I apologize for the back and forth. My previous comment on leveraging existing code didn't mean to changing `cmd/kritis/signer/main.go` actually, and I see that reusing methods there is difficult without a major refactoring.

Can you also fix the Travis tests? You should be able to see the logs.
The GCB tests didn't pass because tests for old modes are failing. I can relay more details if it still breaks later.

mvanholsteijn:
@ooq I opened a new PR (#595) which implements the signer as a stand alone executable.  I processed all of the remarks from this pr in the source code.

I know that this PR looks a bit more complicated, but it is mostly due to the way that the command line processing is done and the fact that the DoCheck and DoCheckAndSign functions needed to be refactored to be used in both the service and the CLI.

I leave it up to you to decide which PR you want to go with.

mvanholsteijn:
PS: I cannot see the logs of the Google Cloud Build Trigger so it is unclear what is wrong.

ooq:
Closing in favor of #595.

PR #590, opened 2020-12-03
Update to k8s v1.16

cjones-shopify:
The goal of this PR is to update Kritis to work with Kubernetes v1.16 (though apparently go modules likes v1.18 just as much!)

There's a few things that I'm uncertain about:

- one of the changes to Kubernetes v1.16 is the removal of `k8s.io/apimachinery/pkg/watch.Until` in  `pkg/kritis/kubernetes/wait.go`. I've replicated what I believe should be identical functionality but I'm relatively new to this part of the Kubernetes API.

Note that this PR includes changes from PR #591, hence the size of it. You can ignore the first two commits (50948e5, 22bdf0c). If the other PR is merged, I will rebase this PR on-top of `master` so it's easier to read.


cjones-shopify:
Closing branch as Go mod won't let me use v0.16.15, so the branch name is inaccurate.

PR #592, opened 2020-12-08
Remove vendor

cjones-shopify:
This PR removes the vendor directory and switches the project to use `go.sum` to ensure that modules aren't tampered with.

Proposing this change in a separate branch rather than my v1.18 branch!


cjones-shopify:
Closing, we will be keeping the vendor directory for consistency with the Kubernetes project.

PR #600, opened 2020-12-16
Add go mod vendor in test.sh to fix module version mismatch

ooq:
Our travis test fails because of `go.mod` and `vendor/modules.txt` mismatch.
The mismatch is introduced by `go get -u gotest.tools/gotestsum` during the testing setup. This command updates `go.mod` but not `vendor/modules.txt`.

Fix is to add `go mod vendor` after downloading `gotestsum`.

### DIFF
--- a/hack/test.sh
+++ b/hack/test.sh
@@ -17,6 +17,7 @@
 if [ ! -x "$(command -v gotestsum)" ] && [ ! -x $GOPATH/bin/gotestsum ]; then
     echo "gotestsum not found, try installing it..."
     go get -u gotest.tools/gotestsum
+    go mod vendor
 fi
 
 set -e


cjones-shopify:
@ooq I think this is because of the change to the `hack/dependencies.sh` script, the Kubernetes one does a lot more slicing and dicing of `go.mod` so that PRs don't get blocked like this. Might I suggest we remove `go mod tidy`?

ooq:
@cjones-shopify Do you mean to remove it `hack/dependencies.sh `? Yesterday I looked, it seemed to be caused by `go get -u gotest.tools/gotestsum` in `./hack/test.sh` updating `go.mod` for `gotestsum`'s dependencies.
Maybe do another `go mod vendor` after installing `gotestsum` would fix this?

ooq:
@cjones-shopify looks like `go mod vendor` does make it happy. Would there be concerns with this approach?

wkozlik (approved):
LGTM

ooq:
Merging now to unblock PRs needing this test to pass. If there is a better fix, let's do so in a follow-up.

PR #601, opened 2020-12-17
Install gotestsum outside of kritis folder

ooq:
As a follow up to #600, this is preferred to avoid mess-ups to developer's local code.


wkozlik (approved):
LGTM

PR #604, opened 2021-01-19
Update sample signing policies

ooq:
Follow up to #603 , updating sample policies that might be used by documentation and tutorials.


wkozlik (approved):
LGTM

